PR #11, opened 2020-05-29
Refactor package.json

oliverlloyd:
- Move version to the top
- Bump src-foundations version
- Added react & react-dom to devDependencies
- Added emotion to peerDependencies


package.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "@guardian/atoms-rendering",
+    "version": "1.2.2",
     "source": "src/index.ts",
     "main": "dist/index.js",
     "module": "dist/index.module.js",
```
gtrufitt, line 6: index.modern.js

PR #20, opened 2020-06-08
remove margin on explainer atom figure

webb04:
Would this work for DCR? it would be useful if atoms came without side padding/margin for AR.

[apps-rendering pull request](https://github.com/guardian/apps-rendering/pull/370)

| Before | After |
| --- | --- |
|   |   |

### DIFF
--- a/src/ExplainerAtom.tsx
+++ b/src/ExplainerAtom.tsx
@@ -22,6 +22,7 @@ const Container = ({
                 padding-bottom: ${space[1]}px;
                 padding-left: ${space[2]}px;
                 padding-right: ${space[2]}px;
+                margin: auto 0;
                 border-top: 1px solid ${text.primary};
                 color: ${text.primary};
                 background: ${neutral[97]};


webb04:
We don't have many global styles in AR but we can override the margin our side. That could be useful, do you use any normalise/reset file at the moment? If we shared similar reset styles it might be easier to share components and not have to include additional styles.

oliverlloyd:
> If we shared similar reset styles it might be easier to share components and not have to include additional styles.

This is a good idea. I think it's worth putting this 'to the group' to get consensus and then having the reset css as part of `example-rendering`.

In atoms, we don't have anything but clearly there's need for one. In DCR, we have [this](https://github.com/guardian/dotcom-rendering/blob/8d0ca732d8223b50360fef05dc6f419a894d3ed4/src/lib/reset-css.js) but I'm not 100% sure that is up to date

webb04:
Ok will do that, I've added the margin removal style to my AR PR so will close this.

PR #23, opened 2020-06-13
Introduce InteractiveAtom

shtukas:
## What does this change?

Implement the InteractiveAtom.

## Images

From storybook:

Screenshot 2020-06-13 at 17 23 03

### DIFF
--- a/src/InteractiveAtom.test.tsx
+++ b/src/InteractiveAtom.test.tsx
@@ -6,8 +6,13 @@ import { InteractiveAtom } from './InteractiveAtom';
 
 describe('InteractiveAtom', () => {
     it('should render', () => {
-        const { getByText } = render(<InteractiveAtom id="123abc" />);
+        const { getByText } = render(
+            <InteractiveAtom
+                id="123abc"
+                url="https://api.nextgen.guardianapps.co.uk/embed/atom/interactive/interactives%2F2020%2F03%2Fcovid-19-uk%2Fdefault"
+            />,
+        );
 
-        expect(getByText('InteractiveAtom')).toBeInTheDocument();
+        // Todo: I don't know which test to put here
     });
 });
--- a/src/InteractiveAtom.tsx
+++ b/src/InteractiveAtom.tsx
@@ -1,7 +1,27 @@
 import React from 'react';
 
+import { css } from 'emotion';
+import { space } from '@guardian/src-foundations';
+import { neutral, text } from '@guardian/src-foundations/palette';
+
 import { InteractiveAtomType } from './types';
 
-export const InteractiveAtom = ({ id }: InteractiveAtomType): JSX.Element => (
-    <div data-atom-id={id}>InteractiveAtom</div>
+export const InteractiveAtom = ({
+    id,
+    url,
+}: InteractiveAtomType): JSX.Element => (
+    <div
+        data-atom-id={id}
+        data-atom-type="interactive"
+        className={css`
+            padding-bottom: ${space[1]}px;
+            padding-left: ${space[2]}px;
+            padding-right: ${space[2]}px;
+            border-top: 1px solid ${text.primary};
+            color: ${text.primary};
+            background: ${neutral[97]};
+        `}
+    >
+        <iframe src={url} frameBorder="0"></iframe>
+    </div>
 );
--- a/src/types.ts
+++ b/src/types.ts
@@ -18,6 +18,7 @@ export type GuideAtomType = {
 
 export type InteractiveAtomType = {
     id: string;
+    url: string;
 };
 
 export type MediaAtomType = {
--- a/src/InteractiveAtom.stories.tsx
+++ b/src/InteractiveAtom.stories.tsx
@@ -8,5 +8,10 @@ export default {
 };
 
 export const DefaultStory = (): JSX.Element => {
-    return <InteractiveAtom id="abc123" />;
+    return (
+        <InteractiveAtom
+            id="abc123"
+            url="https://api.nextgen.guardianapps.co.uk/embed/atom/interactive/interactives%2F2020%2F03%2Fcovid-19-uk%2Fdefault"
+        />
+    );
 };


src/InteractiveAtom.tsx
```diff
@@ -1,7 +1,27 @@
...
+    url,
+}: InteractiveAtomType): JSX.Element => (
+    <div
+        data-atom-id={id}
+        data-atom-type="interactive"
+        className={css`
+            padding-bottom: ${space[1]}px;
+            padding-left: ${space[2]}px;
+            padding-right: ${space[2]}px;
+            border-top: 1px solid ${text.primary};
```
liywjl, line 20: this isnt present on Frontend, the atom seems to automatically come with a grey to border, so no need to add one

src/InteractiveAtom.tsx
```diff
@@ -1,7 +1,27 @@
...
+            color: ${text.primary};
+            background: ${neutral[97]};
```
liywjl, line 22: frontend does not have any background on the atom

liywjl:
I think this PR is ready to review. I am unsure how we can unit test this component as a lot of the logic is all put in the iframe, and as we are virtually rendering the component, unit testing height is not strait forward.

I have added @gtrufitt  to review these changes

src/types.ts
```diff
@@ -39,3 +43,12 @@ export type QuizAtomType = {
...
     id: string;
 };
+
+export type AtomEmbedBlockElement = {
+    _type: string;
+    css: string;
+    js: string;
+    html: string;
+    id: string;
+    url: string;
```
gtrufitt, line 53: Realised we might be missing shouldFence but don't think we need his right now as we'll 'fence' (Stick in an iframe) all atoms.

fixtures/AtomEmbedBlockElement.ts
```diff
@@ -0,0 +1,12 @@
+export const AtomEmbedBlockElement: AtomEmbedBlockElement = {
```
gtrufitt, line 1: This is now InteractiveAtomBlockElement - it'll need find and replace and updating in filenames etc!

```suggestion
export const InteractiveAtomBlockElement: InteractiveAtomBlockElement = {
```

src/InteractiveAtom.tsx
```diff
@@ -1,7 +1,65 @@
...
+        };
+        window.addEventListener('message', setIframeHight);
+        return () => window.removeEventListener('message', setIframeHight);
+    }, []);
+    const markup = unifyPageContent({ js: js, css: css, html: html });
+
+    return (
+        <figure
+            className={figureStyles}
+            data-atom-id={id}
```
gtrufitt, line 53: 💯 This will mean teleporter will still work for atoms 👍

src/lib/unifyPageContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+                <style dangerouslySetInnerHTML={{ __html: css }} />
+            </head>
+            <body>
+                <div dangerouslySetInnerHTML={{ __html: html }} />
+            </body>
+            {/* JS need to load on body render */}
+            <script dangerouslySetInnerHTML={{ __html: js }} />
+            <script src="https://interactive.guim.co.uk/libs/iframe-messenger/iframeMessenger.js"></script>
+            <script>iframeMessenger.enableAutoResize();</script>
+        </html>,
```
gtrufitt, line 25: Rouge comma on the end?

src/lib/unifyPageContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+    css,
+    js,
+    html,
+}: {
+    css: string;
+    js: string;
+    html: string;
+}) =>
+    renderToString(
+        <html>
```
gtrufitt, line 14: Think we're missing a few bits from https://github.com/guardian/frontend/blob/master/common/app/views/fragments/atoms/interactive.scala.html like the doctype and meta. Looks like we also need to load interactive fonts in here.

Basically this file should be equivalent to that existing one.

src/InteractiveAtom.tsx
```diff
@@ -1,7 +1,65 @@
...
+    padding-left: ${space[2]}px;
+    padding-right: ${space[2]}px;
+    color: ${text.primary};
+`;
+
+export const InteractiveAtom = ({
+    id,
+    html,
+    js,
+    css,
```
gtrufitt, line 29: html and css are optional fields

src/lib/unifyPageContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const unifyPageContent = ({
+    css,
+    js,
+    html,
+}: {
+    css: string;
+    js: string;
+    html: string;
+}) =>
```
gtrufitt, line 12: html and css are optional fields

src/lib/unifyPageContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+            <head>
+                <style dangerouslySetInnerHTML={{ __html: css }} />
+            </head>
+            <body>
+                <div dangerouslySetInnerHTML={{ __html: html }} />
+            </body>
+            {/* JS need to load on body render */}
+            <script dangerouslySetInnerHTML={{ __html: js }} />
+            <script src="https://interactive.guim.co.uk/libs/iframe-messenger/iframeMessenger.js"></script>
+            <script>iframeMessenger.enableAutoResize();</script>
```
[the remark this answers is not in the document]
liywjl: I am not well versed in how scala templating builds the iframe. I was following https://github.com/guardian/iframe-messenger#usage, but i can seen in frontend we are "dangerously" inserting JS, is that what we want to do here?
Does frontend run `iframeMessenger.enableAutoResize()`? (I was not able to find its source)

src/lib/unifyPageContent.tsx
liywjl: prettier wanted it :)

src/types.ts
liywjl: I am not quite sure what you mean by `shouldFence`? are you referring to `sandbox` attr in iframe?

src/InteractiveAtom.tsx
```diff
@@ -1,7 +1,65 @@
...
+}: InteractiveAtomType): JSX.Element => {
+    const [height, setHeight] = useState(0);
+    const iframeRef = useRef<HTMLIFrameElement>(null);
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
liywjl: so its been really hard to find a reference point for determining what iframe each message belongs to. I found this example https://github.com/guardian/frontend/blob/6da0b3d8bfd58e8e20f80fc738b070fb23ed154e/static/src/javascripts/projects/facia/modules/ui/snaps.js#L21 I was able to test it and it seemed to work. I have spent sometime on trying to find other approach but this seemed to be the only one that worked well

src/InteractiveAtom.test.tsx
```diff
@@ -6,8 +6,13 @@ import { InteractiveAtom } from './InteractiveAtom';
 
 describe('InteractiveAtom', () => {
     it('should render', () => {
-        const { getByText } = render(<InteractiveAtom id="123abc" />);
+        const { getByText } = render(
+            <InteractiveAtom
+                id="123abc"
+                url="https://api.nextgen.guardianapps.co.uk/embed/atom/interactive/interactives%2F2020%2F03%2Fcovid-19-uk%2Fdefault"
```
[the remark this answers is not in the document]
liywjl: this would be more just testing `unifyPageContent.tsx` than testing `InteractiveAtom`which I am happy to do. I agree though that an integration test makes more sense here

src/InteractiveAtom.tsx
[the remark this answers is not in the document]
gtrufitt: Interactive Atoms are just `srcDoc`ed so you don't need any of this client-side code. `snaps` that you link to there are different things. 

The interactiveAtom html that you generate that is srcDoc'ed (not from CAPI, but the template you set up), and needs to call the interativeResize script - I'll add changes

src/lib/unifyPageContent.tsx
```diff
@@ -0,0 +1,34 @@
...
+                    />
+                    {css && <style dangerouslySetInnerHTML={{ __html: css }} />}
+                </head>
+                <body>
+                    {html && <div dangerouslySetInnerHTML={{ __html: html }} />}
+                </body>
+                {/* JS need to load on body render */}
+                <script dangerouslySetInnerHTML={{ __html: js }} />
+                <script src="https://interactive.guim.co.uk/libs/iframe-messenger/iframeMessenger.js"></script>
+                <script>iframeMessenger.enableAutoResize();</script>
```
gtrufitt, line 31:
```suggestion
<script>
	function resize() {
	    window.frameElement.height = document.body.offsetHeight;
	}
	window.addEventListener('resize', resize);
	resize();
</script>
```

src/types.ts
gtrufitt: No, sorry, I'm referring to this: https://github.com/guardian/frontend/blob/master/common/app/views/fragments/atoms/interactive.scala.html#L36

Basically we 'fence' - iframe - the interativeAtom if on an article page, but otherwise we place it straight into the page. We don't need to worry about it for now though.

src/lib/unifyPageContent.tsx
```diff
@@ -0,0 +1,34 @@
...
+    renderToString(
+        <>
+            {`<!DOCTYPE html>`}
+            <html>
+                <head>
+                    <meta charSet="utf-8" />
+                    <meta
+                        name="viewport"
+                        content="width=device-width,minimum-scale=1,initial-scale=1"
+                    />
```
gtrufitt, line 22: Adding fonts from parent frame

```suggestion
                    />
                    <script>
                    var fonts = [].slice.apply(window.parent.document.styleSheets)
					    .filter(function (sheet) { return sheet.ownerNode.className.indexOf("webfont") > - 1; })
					    .map(function (sheet) { return sheet.ownerNode.textContent; })
					    .join(' ');
					var css = document.createElement('style');
					css.textContent = fonts;
					document.head.appendChild(css);
                    </script>
```

src/types.ts
liywjl: im curious to know were the variable `shouldFence` comes from? should I add it as part of the fixture?

liywjl:
thank you for the clarifications @gtrufitt, I think the PR should be all good now

src/types.ts
gtrufitt: I wouldn't worry for now, it comes via CAPI, but is not in DCRModel yet, so when we need it we can add it.

PR #28, opened 2020-07-16
Q and A Atom

nitro-marky:
## What does this change?

Starting migration of Q and A atom.
Ophan tracking has now been added for EXPAND, LIKE and DISLIKE events.

## Images

Screenshot 2020-07-24 at 10 40 59

Screenshot 2020-07-24 at 10 40 37

Screenshot 2020-07-24 at 10 40 50

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@guardian/atoms-rendering",
-    "version": "1.3.4",
+    "version": "1.3.5",
     "source": "src/index.ts",
     "main": "dist/index.js",
     "module": "dist/index.modern.js",
--- a/package.json
+++ b/package.json
@@ -82,5 +82,8 @@
         "moduleNameMapper": {
             "^@guardian/src-foundations/(.*)$": "@guardian/src-foundations/$1/cjs"
         }
+    },
+    "dependencies": {
+        "@guardian/src-icons": "^2.0.0"
     }
 }
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     },
     "peerDependencies": {
         "@guardian/src-foundations": "^1.1.0",
+        "@guardian/src-icons": "^2.0.0",
         "emotion": "^10.0.27",
         "react": "^16.13.1",
         "react-dom": "^16.13.1"
@@ -82,8 +83,5 @@
         "moduleNameMapper": {
             "^@guardian/src-foundations/(.*)$": "@guardian/src-foundations/$1/cjs"
         }
-    },
-    "dependencies": {
-        "@guardian/src-icons": "^2.0.0"
     }
 }
--- a/src/QandaAtom.stories.tsx
+++ b/src/QandaAtom.stories.tsx
@@ -9,12 +9,29 @@ const html =
 
 const id = '7a6078f6-5a66-4d3e-9339-5610fe320767';
 
+/* const topPaddingText =
+    '<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque lacinia, neque sit amet faucibus pretium, mi lorem laoreet nibh, quis lobortis nisl libero ac dolor. Quisque ut dignissim odio. Mauris velit augue, semper ut dignissim vitae, ullamcorper vitae nulla. Morbi ut dignissim purus, in rutrum dolor. Aenean eu interdum arcu. Cras varius rutrum augue sed semper. Vestibulum ac nisi metus. Sed vitae sapien volutpat est sagittis semper eu eu tortor. Quisque egestas suscipit egestas. Fusce mollis nec lectus ac pulvinar. Donec porta sem ut nunc posuere pretium. Donec quam tortor, euismod eu suscipit a, ullamcorper at neque. Integer vel condimentum libero. Etiam orci massa, pulvinar non tristique id, molestie vulputate est. Proin viverra fringilla est eget vestibulum.</p>';
+
+const bottomPaddingText =
+    '<p>Aenean at molestie nulla. Duis consectetur euismod arcu, quis luctus sem rutrum nec. Vestibulum fringilla, purus ac ultrices laoreet, lorem purus bibendum dolor, et pellentesque libero velit nec lorem. Sed quam tellus, tincidunt eu placerat vitae, tincidunt sit amet ex. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam aliquam erat risus, id tempor nulla consectetur cursus. Nulla sit amet urna laoreet urna consequat cursus eget quis velit. Integer varius non lectus eu commodo. </p>';
+*/
+
 export default {
     title: 'QandaAtom',
     component: QandaAtom,
 };
 
-export const DefaultStory = (): JSX.Element => {
-    //return <QandaAtom id="abc123" />;
+export const LinkStory = (): JSX.Element => {
     return <QandaAtom id={id} title={title} html={html}></QandaAtom>;
+};
+
+const title2 = 'Coronavirus: should everyone be wearing face masks?';
+
+const id2 = '77142d5a-0bf3-4a80-b538-c47a11b57f6a';
+
+const html2 =
+    '<p>Some countries and states have been recommending that everybody wears face masks in indoor settings where social distancing is difficult or impossible. They have been made mandatory on public transport or in shops in many countries.</p><p>According to guidance from the World Health Organization, people over 60 or with health issues should wear a medical-grade mask when they are out and cannot socially distance, while all others should wear a three-layer fabric mask.</p><p>The WHO guidance, announced on 5 June, is a result of research commissioned by the organisation. It is still unknown whether the wearers of masks are protected, say its experts, but the new design it advocates does give protection to other people if properly used.</p><p>The WHO says masks should be made of three layers – with cotton closest to the face, followed by a polypropylene layer and then a synthetic layer that is fluid-resistant. These are no substitute for physical distancing and hand hygiene, it says, but should be worn in situations where distancing is difficult, such as on public transport and at mass demonstrations.</p><p>The WHO has been reluctant to commit to recommending face coverings, firstly because the evidence on whether they offer any protection to the public is limited and – more importantly – because it was afraid it would lead to shortages of medical-grade masks for health workers.</p><p><b>&nbsp;<a href="https://www.theguardian.com/profile/sarahboseley" data-link-name="in body link" class="u-underline">Sarah Boseley</a></b>&nbsp;<i>Health editor</i></p>';
+
+export const AuthorStory = (): JSX.Element => {
+    return <QandaAtom id={id2} title={title2} html={html2}></QandaAtom>;
 };
--- a/src/QandaAtom.stories.tsx
+++ b/src/QandaAtom.stories.tsx
@@ -9,20 +9,21 @@ const html =
 
 const id = '7a6078f6-5a66-4d3e-9339-5610fe320767';
 
-/* const topPaddingText =
-    '<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque lacinia, neque sit amet faucibus pretium, mi lorem laoreet nibh, quis lobortis nisl libero ac dolor. Quisque ut dignissim odio. Mauris velit augue, semper ut dignissim vitae, ullamcorper vitae nulla. Morbi ut dignissim purus, in rutrum dolor. Aenean eu interdum arcu. Cras varius rutrum augue sed semper. Vestibulum ac nisi metus. Sed vitae sapien volutpat est sagittis semper eu eu tortor. Quisque egestas suscipit egestas. Fusce mollis nec lectus ac pulvinar. Donec porta sem ut nunc posuere pretium. Donec quam tortor, euismod eu suscipit a, ullamcorper at neque. Integer vel condimentum libero. Etiam orci massa, pulvinar non tristique id, molestie vulputate est. Proin viverra fringilla est eget vestibulum.</p>';
-
-const bottomPaddingText =
-    '<p>Aenean at molestie nulla. Duis consectetur euismod arcu, quis luctus sem rutrum nec. Vestibulum fringilla, purus ac ultrices laoreet, lorem purus bibendum dolor, et pellentesque libero velit nec lorem. Sed quam tellus, tincidunt eu placerat vitae, tincidunt sit amet ex. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam aliquam erat risus, id tempor nulla consectetur cursus. Nulla sit amet urna laoreet urna consequat cursus eget quis velit. Integer varius non lectus eu commodo. </p>';
-*/
-
 export default {
     title: 'QandaAtom',
     component: QandaAtom,
 };
 
-export const LinkStory = (): JSX.Element => {
-    return <QandaAtom id={id} title={title} html={html}></QandaAtom>;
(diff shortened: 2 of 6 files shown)


nitro-marky:
[Atom: QandaAtom](https://trello.com/c/He2VyjwF/1731-atom-qandaatom)

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+                        className={buttonRound}
+                        value="dislike"
+                        aria-label="No"
+                        onClick={() => {
+                            dislikeHandler();
+                            setFeedbackVisibility(true);
+                        }}
+                    >
+                        <svg
+                            className={thumbIcon + ' ' + dislikeThumb}
```
gtrufitt, line 405: Are combining classes here? You can use `cx` from emotion

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+                aria-hidden="true"
+            >
+                <span className={'is-on ' + iconSpacing}>
+                    <svg
+                        className={iconStyling}
+                        width="18px"
+                        height="18px"
+                        viewBox="0 0 18 18"
+                    >
+                        <path d="M8.2 0h1.6l.4 7.8 7.8.4v1.6l-7.8.4-.4 7.8H8.2l-.4-7.8L0 9.8V8.2l7.8-.4.4-7.8z"></path>
```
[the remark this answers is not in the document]
gtrufitt: Same witht the other icons prefereably (it'll mean if we ever change icons, we can just bump design system, theoretically)

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+                            <path
+                                fill="#FFF"
+                                d="M33.78 22.437l-4.228 13.98L27.93 37.5 5.062 34.14V15.503l7.8-1.517L24.354 2.5h1.624L28.9 5.426l-4.548 8.67h.107l10.477 1.31"
+                            ></path>
+                        </svg>
+                    </button>
+                </div>
+            </div>
+            <div
+                className={'feedback' + ' ' + footerFeedback}
```
gtrufitt, line 419: What's this 'feedback' class for?

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+                    line-height: 22px;
+                `}
+            >
+                {title}
+            </h4>
+            <span
+                className={showHideStyling}
+                onClick={() => {
+                    if (!hasBeenExpanded) {
+                        expandHandler();
```
gtrufitt, line 286: I'd check that this handler isn't send multiple times - you may need to put it inside `useEffect` with the `hasBeenExpanded` state as the conditional source prop https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect (maybe not as in onClick but this has been be a source of bugs in my code before!)

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+                        className={iconStyling}
+                        width="18px"
+                        height="18px"
+                        viewBox="0 0 18 18"
+                    >
+                        <path d="M8.2 0h1.6l.4 7.8 7.8.4v1.6l-7.8.4-.4 7.8H8.2l-.4-7.8L0 9.8V8.2l7.8-.4.4-7.8z"></path>
+                    </svg>
+                    Show
+                </span>
+                <span className={'is-off ' + iconSpacing}>
```
gtrufitt, line 303: Not sure where the `is-off` and `is-on` are used and wether you need them?

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+    color: white;
+    border-style: hidden;
+    border-radius: 100%;
+    margin: 0 0 0 5px;
+    padding: 0;
+    width: 28px;
+    height: 28px;
+
+    :hover {
+        background: ${error[400]};
```
gtrufitt, line 142: Semantically, it doesn't seem quite right to use the `error` named colour here, because if that changes in the design system then it's likely you didn't want it to change here. If the colour is not in the palette (it should be though, maybe it's news red?), then I would use the hex value instead.

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+    }
+`;
+
+const figureStyling = css`
+    display: block;
+    position: relative;
+    margin-bottom: 0.75rem;
+    margin-top: 1rem;
+
+    details[open] .is-on {
```
gtrufitt, line 90: Ah ok, here is the `is-on`, `is-off` - I would do the conditonal classes with js teneary operator here as you have the `expanded` state anyway (it's quite nice using the css classes, but it will be easier for future maintainers to reason with if we use one way to define and read state)

So below instead of `<span className={'is-on ' + iconSpacing}>` use `hasBeenExpanded && <span className={iconSpacing}>`

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+        margin: 0 0 0.75rem;
+        padding: 0;
+    }
+
+    li {
+        margin-bottom: 0.375rem;
+        padding-left: 1.25rem;
+    }
+
+    li:before {
```
gtrufitt, line 53: Hmm, we seem to be duplicating this into here, what we're doing in dotcom rendering. not sure what the solution is off the top of my head!

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
-
 import { QandaAtomType } from './types';
+import { css } from 'emotion';
+import { neutral, error } from '@guardian/src-foundations/palette';
+import { headline, textSans, body } from '@guardian/src-foundations/typography';
+
+const { useState } = React;
+
+// CSS
+const detailStyling = css`
```
gtrufitt, line 10: Can I suggest that this is targeted to the body rather than the detail element - so that you're more closely targeting the dangerouslySetHtml part (where you have to style elements) vs the outer elements  you have control over that you can style with className more easily

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
 import React from 'react';
-
 import { QandaAtomType } from './types';
+import { css } from 'emotion';
+import { neutral, error } from '@guardian/src-foundations/palette';
+import { headline, textSans, body } from '@guardian/src-foundations/typography';
+
+const { useState } = React;
```
gtrufitt, line 7: Should be moved to line 1 to be import React, {useState} from 'react';

src/QandaAtom.tsx
gtrufitt: The reason I ask is that it's often an anti-pattern in DCR to see string classes like this

src/QandaAtom.tsx
gtrufitt: https://github.com/guardian/atoms-rendering/pull/28#discussion_r460816811

src/QandaAtom.stories.tsx
```diff
@@ -2,11 +2,122 @@ import React from 'react';
...
...href="https://www.who.int/emergencies/diseases/novel-coronavirus-2019/advice-for-public">advises</a>&nbsp;people to:</p> <ul> <li>Frequently wash their hands with an alcohol-based hand rub or warm water and soap</li> <li>Cover their mouth and nose with a flexed elbow or tissue when sneezing or coughing</li> <li>Avoid close contact with anyone who has a fever or cough</li> <li>Seek early medical help if they have a fever, cough and difficulty breathing, and share their travel history with healthcare providers</li> <li>Advice about face masks varies. Wearing them while out and about may offer some protection against both spreading and catching the virus via coughs and sneezes, but it is not a cast-iron guarantee of protection</li> </ul> <p>Many countries are now enforcing or recommending curfews or lockdowns. <b>Check with your local authorities for up-to-date information about the situation in your area.</b>&nbsp;</p> <p>In the UK, NHS advice is that anyone with symptoms should&nbsp;<b>stay at home for at least 7 days</b>.</p> <p>If you live with other people,&nbsp;<b>they should stay at home for at least 14 days</b>, to avoid spreading the infection outside the home.</p> <p></p> ';
```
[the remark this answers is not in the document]
nitro-marky: Not sure where they came from, I've removed them.

src/QandaAtom.tsx
```diff
@@ -1,7 +1,445 @@
...
+}: {
+    html: string;
+    image?: string;
+    credit?: string;
+}) => (
+    <div>
+        {image && <img className={imageStyling} src={image} alt="" />}
+        <div
+            className={css`
+                ${body.medium()}
```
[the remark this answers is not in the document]
nitro-marky: This has now been moved into bodyStyling

src/QandaAtom.tsx
nitro-marky: I think that was left over from some tests I was doing, removed now

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+    :focus {
+        border: none;
+    }
+`;
+
+const imageStyling = css`
+    float: left;
+    margin-right: 16px;
+    margin-bottom: 6px;
+    width: 100px !important;
```
gtrufitt, line 145: `!important` is often a code smell - what are you having to override here (and other usages)?

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+
+const Summary = ({
+    title,
+    expandHandler,
+}: {
+    title: string;
+    expandHandler: () => void;
+}) => {
+    const [hasBeenExpanded, setExpandStatus] = useState(false);
+    const [expandEventSent, setExpandEventFired] = useState(false);
```
gtrufitt, line 260: Convention suggests that the `set` method should match the name of the state variable ie:`hasBeenExpanded, setHasBeenExpanded` - which makes it easier to reason with in the code (same for expandEvent`

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+    return (
+        <summary className={summaryStyling}>
+            <span
+                className={css`
+                    color: #e00000;
```
gtrufitt, line 265: Should this be a palette colour?

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+                    ${headline.xxxsmall({
+                        fontWeight: 'medium',
+                    })};
+                    margin: 0;
+                    line-height: 22px;
+                `}
+            >
+                {title}
+            </h4>
+            <span
```
gtrufitt, line 286: Worth double check the keyboard navigation, I assume the 'details' element allows tabbing / spacebar to open, but as we're aria-hidden on this we should make sure it is still accessible to screen reader /keyboard users

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+            }}
+        />
+        {credit && <Credit credit={credit} />}
+    </div>
+);
+
+const Credit = ({ credit }: { credit: string }) => (
+    <div className={creditStyling}>
+        <span className={creditSpanStyling}>
+            <svg
```
gtrufitt, line 345: No Source icon for this?

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+                        className={buttonRound}
+                        value="like"
+                        aria-label="Yes"
+                        data-testid="like"
+                        onClick={() => {
+                            likeHandler();
+                            setFeedbackVisibility(true);
+                        }}
+                    >
+                        <svg
```
gtrufitt, line 381: Source icon?

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+    likeHandler,
+    dislikeHandler,
+}: {
+    likeHandler: () => void;
+    dislikeHandler: () => void;
+}) => {
+    const [showFeedback, setFeedbackVisibility] = useState(false);
+    return (
+        <footer className={footerStyling}>
+            <div hidden={showFeedback ? true : false}>
```
gtrufitt, line 368:
```suggestion
            <div hidden={showFeedback}>
```

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
-export const QandaAtom = ({ id }: QandaAtomType): JSX.Element => (
-    <div data-atom-id={id}>QandaAtom</div>
+const Footer = ({
+    likeHandler,
+    dislikeHandler,
+}: {
+    likeHandler: () => void;
+    dislikeHandler: () => void;
+}) => {
+    const [showFeedback, setFeedbackVisibility] = useState(false);
```
gtrufitt, line 365: Same with state naming

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+                                d="M33.78 22.437l-4.228 13.98L27.93 37.5 5.062 34.14V15.503l7.8-1.517L24.354 2.5h1.624L28.9 5.426l-4.548 8.67h.107l10.477 1.31"
+                            ></path>
+                        </svg>
+                    </button>
+                </div>
+            </div>
+            <div
+                className={footerFeedback}
+                aria-live="polite"
+                hidden={showFeedback ? false : true}
```
gtrufitt, line 426:
```suggestion
                hidden={!showFeedback}
```

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+                        aria-label="No"
+                        data-testid="dislike"
+                        onClick={() => {
+                            dislikeHandler();
+                            setFeedbackVisibility(true);
+                        }}
+                    >
+                        <svg
+                            className={
+                                thumbIcon +
```
gtrufitt, line 405: I'd pull this into a const

src/QandaAtom.tsx
```diff
@@ -1,7 +1,450 @@
...
+                                ' ' +
+                                css`
+                                    transform: rotate(180deg);
+                                `
+                            }
+                            width="40px"
+                            height="40px"
+                            viewBox="0 0 40 40"
+                        >
+                            <path
```
gtrufitt, line 415: Source icons?

package.json
```diff
@@ -82,5 +82,8 @@
         "moduleNameMapper": {
             "^@guardian/src-foundations/(.*)$": "@guardian/src-foundations/$1/cjs"
         }
+    },
+    "dependencies": {
```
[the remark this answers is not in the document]
gtrufitt: Most `source` packages peer dependencies because we know that the platform will have them installed and we don't need to bundle them with our module

src/QandaAtom.tsx
nitro-marky: The current icon set doesn't have the thumbs one

src/QandaAtom.tsx
nitro-marky: Not for thumbs at the moment

gtrufitt: Cool, leave as is, we can summon @guardian/design-system to see if thumbs up and thumbs down can be added in the future to refactor in.

src/QandaAtom.tsx
```diff
@@ -138,12 +138,13 @@ const buttonRound = css`
     }
 `;
 
+// The important is here to override the default image style which defaults to 100% width
```
gtrufitt, line 141: So I understand this, where is that default style coming from? It should be that specific styles always override default styles from the cascade, we should never need to use `!important`

nitro-marky: I think it's coming from DCR, I haven't added any `width: 100%` styles anywhere in atoms-rendering

PR #31, opened 2020-07-30
1.5.0 Version bump

nitro-marky:
## What does this change?

This is the basis for atoms-rendering version 1.5.0. This version now includes Q and A atoms

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
         "@babel/preset-react": "^7.9.4",
         "@babel/preset-typescript": "^7.9.0",
         "@guardian/src-foundations": "^1.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -83,5 +83,8 @@
         "moduleNameMapper": {
             "^@guardian/src-foundations/(.*)$": "@guardian/src-foundations/$1/cjs"
         }
+    },
+    "dependencies": {
+        "@guardian/src-icons": "1.9.1"
```
gtrufitt, line 88: This should be a devDependency actually!

gtrufitt: As well as per

PR #33, opened 2020-08-05
Guide Atom

nitro-marky:
## What does this change?
This adds the guide atom to atoms-rendering and includes Storybook and unit tests. Pillar colours are also now working. 

Elements like the summary box (with expansion buttons) and the feedback buttons have been abstracted out into SharedElement.

## Images
Screenshot 2020-08-05 at 16 31 40

Screenshot 2020-08-05 at 16 24 07

Screenshot 2020-08-05 at 16 24 21

Screenshot 2020-08-05 at 16 25 36

Screenshot 2020-08-10 at 12 10 55


src/GuideAtom.tsx
```diff
@@ -1,33 +1,206 @@
...
+        <details
+            className={detailStyling}
+            data-atom-id={id}
+            data-atom-type="guide"
+        >
+            <Summary
+                sectionTitle={'Quick Guide'}
+                title={title}
+                expandHandler={expandHandler}
+            ></Summary>
```
jfsoul, line 149: I think this could be self-closing?

src/SharedElements.tsx
```diff
@@ -0,0 +1,237 @@
+import React, { useState } from 'react';
+import { css, cx } from 'emotion';
+import { textSans, headline, body } from '@guardian/src-foundations/typography';
+import { news, neutral } from '@guardian/src-foundations/palette';
+import { SvgMinus, SvgPlus } from '@guardian/src-icons';
+
+/*
+    Elements that are used on multiple atoms (such as the feedback buttons) 
+    can be imported from here
```
jfsoul, line 9: I would have these in separate files, perhaps under a `components` folder or something.

src/SharedElements.tsx
```diff
@@ -0,0 +1,237 @@
...
+*/
+
+/// SUMMARY ELEMENT
+
+const showHideStyling = css`
+    background: ${neutral[7]};
+    color: ${neutral[100]};
+    height: 2rem;
+    font-size: 13px;
```
jfsoul, line 18: is this in the design system?

src/SharedElements.tsx
```diff
@@ -0,0 +1,237 @@
...
+                    ${body.medium({
+                        lineHeight: 'tight',
+                        fontWeight: 'bold',
+                    })};
+                `}
+            >
+                {sectionTitle}
+            </span>
+            <h4
+                className={css`
```
jfsoul, line 91: Maybe don't inline this css? It's nice to give them a name for readability here and to be able to reference back when they're rendered in the browser.

jfsoul: (or the one just above)

src/SharedElements.tsx
```diff
@@ -0,0 +1,237 @@
...
+            <span className={showHideStyling}>
+                {!hasBeenExpanded && (
+                    <span className={iconSpacing}>
+                        <span className={plusStyling}>
+                            <SvgPlus />
+                        </span>
+                        Show
+                    </span>
+                )}
+                {hasBeenExpanded && (
```
jfsoul, line 110: you could do this with a ternary operator i.e. `hasBeenExpanded ? <expandedMarkup> : <notExpandedMarkup>;`

src/SharedElements.tsx
```diff
@@ -0,0 +1,237 @@
...
+};
+
+export const Footer = ({
+    likeHandler,
+    dislikeHandler,
+}: {
+    likeHandler: () => void;
+    dislikeHandler: () => void;
+}): JSX.Element => {
+    const [showFeedback, setShowFeedback] = useState(false);
```
jfsoul, line 181: Maybe rename this `showThankYou` or flip it round so that it actually is `showFeedback` (if by "feedback" we mean the thumbs up/down buttons).

src/GuideAtom.test.tsx
```diff
@@ -1,21 +1,105 @@
...
+                html="<p><b>Happy Valley&nbsp;</b></p><p><b>11.45</b> Happy Good Guys <b>12.15</b> Salto Olimpico <b>12.45</b> Seize The Spirit <b>1.15</b> Allied Agility <b>1.45 </b>Hero Time <b>2.15</b> Simply Fluke <b>2.45</b> Brave King <b>3.15</b> Golden Dash <b>3.50</b> This Is Class</p>"
+                credit=""
+                likeHandler={() => {
+                    return null;
+                }}
+                dislikeHandler={() => {
+                    return null;
+                }}
+                expandHandler={() => {
+                    return null;
```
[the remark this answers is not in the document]
jfsoul: I guess I would expect it to be called something like `expandCallback` which implies we take some action internally to expand something and then call this function as a secondary action.

jfsoul:
Nice work on this! 

A thought for next time: it's nice to try and break down PRs into the smallest releasable chunks so that reviewing is easier, bugs are easier to isolate, feedback is faster etc. I mention because this might've been a decent candidate to split into 2 parts: refactor shared components out of the Q and A Atom, and then introduce the Guide Atom in a followup.

src/GuideAtom.stories.tsx
```diff
@@ -16,6 +16,82 @@ export const DefaultStory = (): JSX.Element => {
...
+                return null;
+            }}
+            expandHandler={() => {
+                return null;
+            }}
+        />
+    );
+};
+
+export const ImageStory = (): JSX.Element => {
```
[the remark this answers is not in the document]
nitro-marky: These seem to be the main styles used inside of guide atoms, so I thought it would be useful to have an example for each. Mainly to check that list styles or ul and ol don't interfere with each other. It's also useful to check one with an image to make sure the logic works correctly in terms of spacing there.

src/GuideAtom.tsx
```diff
@@ -1,33 +1,206 @@
...
+import { SvgInfo } from '@guardian/src-icons';
+import { Footer, Summary } from './SharedElements';
 
-const Body = ({ html }: { html: string }) => (
-    <div
-        dangerouslySetInnerHTML={{
-            __html: html,
-        }}
-    />
+// CSS
```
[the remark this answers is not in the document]
nitro-marky: I've removed it, I included it just to demarcate the styling section and the component section.

src/GuideAtom.tsx
```diff
@@ -1,33 +1,206 @@
...
+    border-top: 13px solid black;
+    border-left: 0;
+    border-right: 0;
+    border-bottom: 0;
+    position: relative;
+    summary {
+        list-style: none;
+        margin: 0 0 16px;
+    }
+    summary::-webkit-details-marker {
```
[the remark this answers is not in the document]
nitro-marky: Part of the cross-browser tools to stop the summary from getting an outline when focused

src/QandaAtom.tsx
```diff
@@ -48,25 +48,25 @@ const bodyStyling = css`
     p {
```
gtrufitt, line 48: I suspect you want to refactor the Q&A atom and move the footer and summary components to be shared

src/components/Summary.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <span
+                className={
+                    showHideStyling +
+                    ' ' +
+                    css`
+                        :hover {
+                            background: ${linkColourStyleHover};
+                        }
+                    `
+                }
```
gtrufitt, line 138: I don't think this needs to be a seperate style here, can it not just live inside the `showHideStyling(pillar)`?

src/components/Summary.tsx
```diff
@@ -0,0 +1,158 @@
...
+        case 'labs':
+            linkColourStyleHover = labs[400];
+            titleColour = labs[400];
+            break;
+        case 'news':
+            linkColourStyleHover = news[400];
+            titleColour = news[400];
+            break;
+    }
+}
```
gtrufitt, line 93: Setting `let` variables can make it difficult to follow the code, so using a more functional approach is favoured where we `return` the colour (especially as they're all the same):

```suggestion
function getPillarHighlight(pillar: Pillar) { 
    switch (pillar) {
        case 'opinion':
            return opinion[400];
        case 'sport':
			return sport[400];
        case 'culture':
            return culture[400];
        case 'lifestyle':
            return lifestyle[400];
        case 'labs':
			return labs[400];
        case 'news':
        default:
			return news[400];
    }
}
```

and use it in the css like: 

```
background: ${getPillarHighlight(pillar)};
```

src/components/Summary.tsx
```diff
@@ -0,0 +1,158 @@
...
+    sport,
+    lifestyle,
+    labs,
+} from '@guardian/src-foundations/palette/';
+import { SvgMinus, SvgPlus } from '@guardian/src-icons';
+
+/// SUMMARY ELEMENT
+let linkColourStyleHover = '';
+let titleColour = '';
+
```
gtrufitt, line 18: As per comment below, having `let` vars can be a code smell

```suggestion
```

src/GuideAtom.stories.tsx
```diff
@@ -16,6 +16,112 @@ export const DefaultStory = (): JSX.Element => {
             title="Wednesday's Hong Kong tips"
             html="<p><b>Happy Valley&nbsp;</b></p><p><b>11.45</b> Happy Good Guys <b>12.15</b> Salto Olimpico <b>12.45</b> Seize The Spirit <b>1.15</b> Allied Agility <b>1.45 </b>Hero Time <b>2.15</b> Simply Fluke <b>2.45</b> Brave King <b>3.15</b> Golden Dash <b>3.50</b> This Is Class</p>"
             credit=""
```
gtrufitt, line 18: Your stories are all of the closed component, which I assume isn't what you want. You probably need to add an optional 'expanded' prop on the component that is used as the default state of the exapanded

src/GuideAtom.tsx
```diff
@@ -1,33 +1,255 @@
...
+        width: 0.75rem;
+        margin-right: 0.5rem;
+        background-color: ${neutral[86]};
+        margin-left: -1.25rem;
+    }
+
+    b {
+        font-weight: bold;
+    }
+`;
```
gtrufitt, line 94: This body styling looks familiar to the other atom, I imagine you're going to be duplicating it again, so it's worth working out how to share it. I have a feeling these atoms are going to be very much 'containers' for the abstracted body, footer, summary etc inside?

src/GuideAtom.tsx
```diff
@@ -1,33 +1,255 @@
...
+    pillar,
+    expandCallback,
+}: {
+    id: string;
+    title: string;
+    pillar: string;
+    children: React.ReactNode;
+    expandCallback: () => void;
+}) => (
+    <div className={containerStyling} data-atom-id={id} data-atom-type="guide">
```
gtrufitt, line 136: You have data-atom-type here and on detials, does it need to be on both?

src/GuideAtom.tsx
```diff
@@ -1,33 +1,255 @@
...
+        </div>
+    );
+};
+
+const Credit = ({ credit }: { credit: string }) => (
+    <div className={creditStyling}>
+        <SvgInfo />
+        {credit}
+    </div>
 );
```
gtrufitt, line 201: This abstraction probably is just an unneeded bit of cognitive load, so worth inlining. 

```suggestion
            {credit && (<div className={creditStyling}><SvgInfo />{credit}</div>)}
        </div>
    );
};
```

src/components/Summary.tsx
gtrufitt: I think you probably want to pull it into a `lib` function as well as you've got a few places you have this pattern that you'll want to remove

src/QandaAtom.tsx
nitro-marky: Yeah, I want to get the guide done first and merged then refactor Q and A

gtrufitt: OK! Yes, take my other comments with that in mind then 👍

src/GuideAtom.tsx
```diff
@@ -1,33 +1,224 @@
...
+        )
+        13;
+    border-top: 13px solid black;
+    position: relative;
+    summary {
+        list-style: none;
+        margin: 0 0 16px;
+    }
+
+    summary::-webkit-details-marker {
```
gtrufitt, line 38:
```suggestion
	// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Customizing_the_disclosure_widget
    summary::-webkit-details-marker {
```

src/GuideAtom.tsx
```diff
@@ -1,33 +1,224 @@
...
+        height: 0.75rem;
+        width: 0.75rem;
+        margin-right: 0.5rem;
+        background-color: ${neutral[86]};
+        margin-left: -1.25rem;
+    }
+
+    b {
+        font-weight: bold;
+    }
```
gtrufitt, line 84: Is this needed? Does the user agent stylesheet get overriden in DCR?

src/GuideAtom.tsx
```diff
@@ -1,33 +1,224 @@
...
+                            text-decoration: none;
+                            border-bottom: 0.0625rem solid ${neutral[86]};
+                            transition: border-color 0.15s ease-out;
+                        }
+
+                        a:hover {
+                            border-bottom: solid 0.0625rem
+                                ${GetPillarColour400(pillar)};
+                        }
+                    `,
```
gtrufitt, line 180: This is a large block of CSS, I'd assign in to a const

src/lib/PillarColours.tsx
```diff
@@ -0,0 +1,44 @@
...
+            return culture[300];
+        case 'lifestyle':
+            return lifestyle[300];
+        case 'labs':
+            return labs[300];
+        case 'news':
+        default:
+            return news[300];
+    }
+}
```
gtrufitt, line 44: Can reduce this to a single function with a parameter for the colour value.

src/GuideAtom.tsx
nitro-marky: Yeah, bold styling gets overridden without this.

gtrufitt: I think you likely want to use the font weight here https://theguardian.design/2a1e5182b/p/930d69-overview/t/288fd4 so - 700? Otherwise you might get faux bold rather than use the actual bold typeface.

PR #35, opened 2020-08-12
Introduce Profile Atom

shtukas:
## What does this change?

Introduce the `Profile Atom` based on the amazing recent work by @nitro-marky on the Guide Atom.

Screenshot 2020-08-12 at 08 51 04

Screenshot 2020-08-12 at 08 51 11


shtukas:
+1

PR #36, opened 2020-08-14
Timeline Atom

nitro-marky:
## What does this change?
This adds the timeline atom to atoms-rendering. It support formatted timelines and timelines created as a block of text without bullet points.

## Images

Screenshot 2020-08-14 at 14 53 19

Screenshot 2020-08-14 at 14 52 54

Screenshot 2020-08-14 at 14 53 38

Screenshot 2020-08-14 at 14 53 54


src/components/Body.tsx
```diff
@@ -0,0 +1,124 @@
...
+    image,
+    credit,
+    pillar,
+}: {
+    html: string;
+    image?: string;
+    credit?: string;
+    pillar: string;
+}): JSX.Element => {
+    const linkStyling = css`
```
gtrufitt, line 91: Probably worth putting this outside of the function. Really like splitting up of `linkStyling` and `bodyStyling` though!

src/TimelineAtom.tsx
```diff
@@ -1,7 +1,115 @@
...
+            title={title}
+            expandCallback={expandCallback}
+        >
+            {description && <Body html={description} pillar={pillar} />}
+            {events && <TimelineContents events={events} pillar={pillar} />}
+            <Footer
+                pillar={pillar}
+                dislikeHandler={dislikeHandler}
+                likeHandler={likeHandler}
+            ></Footer>
```
gtrufitt, line 112:
```suggestion
            <Footer
                pillar={pillar}
                dislikeHandler={dislikeHandler}
                likeHandler={likeHandler}
            />
```

src/components/Body.tsx
nitro-marky: I put it there as it needs to use the pillar parameter passed to the component. Is there a way you recommend to make it accessible outside?

gtrufitt: You can do ` const linkStyling = (pillar) => css` and call it in the `className={linkStyling(pillar)}`

nitro-marky: Cool, thanks

src/TimelineAtom.tsx
```diff
@@ -1,7 +1,115 @@
 import React from 'react';
+import { Container } from './components/Container';
+import { Footer } from './components/Footer';
+import { Body } from './components/Body';
```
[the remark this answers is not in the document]
nitro-marky: I've changed the components folder to expandableAtom, do you think the individual elements need changing?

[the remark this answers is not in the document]
gtrufitt: 👍 Nope, think the folder gives enough context!

PR #44, opened 2020-08-24
Chart Atom

nitro-marky:
## What does this change?
A couple of minor changes from Pascal's start. Embedding urls needs some work and is causing problem with iframe height and resizing.

Turned out to be a way bigger challenge getting the resizing to work properly!!!
Screenshot 2020-09-07 at 15 59 26


shtukas (approved):

src/ChartAtom.tsx
```diff
@@ -1,23 +1,40 @@
...
+            data-atom-id={id}
+            data-testid="chart"
+            data-atom-type="chart"
+            data-snippet-type="chart"
+            className={css`
+                padding-bottom: ${space[1]}px;
+                padding-left: ${space[2]}px;
+                padding-right: ${space[2]}px;
+                color: ${text.primary};
+                margin-bottom: ${space[2]}px;
```
gtrufitt, line 28: Are the paddings and margins definitely required for DCR? Certainly the right and left paddings should be handled by the article gutters I think?

PR #45, opened 2020-08-25
Add Quiz Atom

amyhughes:
## What does this change?
Here we add support for a `QuizAtom`, designed to render like the existing quiz atom found for example on [this](https://www.theguardian.com/football/that-1980s-sports-blog/2020/jun/12/sports-quiz-football-in-the-1980s) page.

[This frontend PR](https://github.com/guardian/frontend/pull/22931) shows the proposed data to be passed to DCR for the element, and a story has been added to render the atom for the JSON returned when running the `add-quiz-atom-block-element` frontend branch locally for the [example page](https://www.theguardian.com/football/that-1980s-sports-blog/2020/jun/12/sports-quiz-football-in-the-1980s).

TODO: The existing implementation [supports posting results](https://github.com/guardian/frontend/blob/master/common/app/quiz/package.scala#L67), I need to add this functionality to the implementation (if anyone has a link to an example of a quiz where you post the results it would be much appreciated!).

## Images
For this example, the quiz looks and behaves (mostly) as expected:

### QuizAtom story (on hover)
image

### QuizAtom story (on click)
image

### Existing frontend rendering
image

## How to test

On this branch, run `yarn storybook` and look at the Quiz Atom > Default Story


src/QuizAtom.tsx
```diff
@@ -1,7 +1,317 @@
...
-import { QuizAtomType } from './types';
+import {
+    QuizAtomType,
+    Question as QuestionType,
+    Answer as AnswerType,
+} from './types';
 
-export const QuizAtom = ({ id }: QuizAtomType): JSX.Element => (
-    <div data-atom-id={id}>QuizAtom</div>
+const enum AnswerState {
```
jfsoul, line 13: I like this way of encapsulating the states 👍

amyhughes: Haha thanks Jon, both you and Nic looked slightly horrified by how this was done before! It's much nicer now for sure.

.babelrc.js
```diff
@@ -4,6 +4,15 @@ module.exports = {
         '@babel/preset-typescript',
         '@babel/preset-react',
     ],
+    plugins: [
+        'const-enum',
```
gtrufitt, line 8: It'd be great to add a link in a comment to https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats as I wasn't sure why this was needed :)

src/QuizAtom.test.tsx
```diff
@@ -6,8 +6,8 @@ import { QuizAtom } from './QuizAtom';
 
 describe('QuizAtom', () => {
     it('should render', () => {
-        const { getByText } = render(<QuizAtom id="123abc" />);
+        const { getByText } = render(<QuizAtom id="123abc" questions={[]} />);
```
gtrufitt, line 9: The react testing library supports `fireEvent` for clicking, in case that's useful for this example! https://testing-library.com/docs/react-testing-library/cheatsheet

Probably a light test, as Cypress for an integration test in DCR is probably a nicer option for testing user behaviour

src/QuizAtom.tsx
```diff
@@ -1,7 +1,322 @@
...
+                        {number + '.'}
+                    </div>
+                    <legend
+                        className={css`
+                            margin-bottom: 12px;
+                        `}
+                    >
+                        {text}
+                    </legend>
+                </div>
```
gtrufitt, line 221: classNames assigned to classes can be useful in places like this for two reasons:

A classname on the outer div can 'name' the block of html so you're able to style with one level deep element styling and have some context:

```
const questionStyle = css`
	display: flex;
	
	> div {
		padding-right: 0.5ch;
	}

	> legend {
		 margin-bottom: 12px;
	}
`
```

I then have a little more context what this part of the component is.

Also, when developing it's useful to have named classNames (not so much in atoms rendering as I am not sure it is turned on!) in DCR at least it adds the className to the end of the class in dev mode (not prod) so it's much easier to find where your element is and where to style.

This is all slightly stylistic but I've found shallow direct-child styling against a contextual className can be the nice mix of devEx and not bloating CSS.

src/QuizAtom.tsx
```diff
@@ -1,7 +1,322 @@
...
+
+const IconWrapper = (el: JSX.Element) => <div className={iconStyle}>{el}</div>;
+
+export const QuizAtom = ({ id, questions }: QuizAtomType): JSX.Element => (
+    <div data-atom-id={id}>
+        <form>
+            {questions.map((question) => (
+                <Question
+                    key={question.id}
+                    number={questions.indexOf(question) + 1}
```
gtrufitt, line 175: Interesting! `map` does provide index as a second param `map((questions, idx`), which might be a bit easier to reason with? Took me a beat to understand what was going as there's a bit of misdirection going on.

src/QuizAtom.tsx
```diff
@@ -1,7 +1,322 @@
...
+`;
+
+const fieldsetStyle = css`
+    margin-bottom: 12px;
+    border: 0px;
+    padding: 0px;
+`;
+
+// Will appear on screen readers
+const visuallyHidden = css`
```
gtrufitt, line 143: Source provides this as a helper `import { visuallyHidden } from "@guardian/src-foundations/accessibility"`

src/QuizAtom.tsx
```diff
@@ -1,7 +1,322 @@
...
+                    <legend
+                        className={css`
+                            margin-bottom: 12px;
+                        `}
+                    >
+                        {text}
+                    </legend>
+                </div>
+                {imageUrl && (
+                    <img className={fullWidthStyle} src={imageUrl}></img>
```
gtrufitt, line 223: Do you get image width and height at all? Useful to have them for rendering but no worries if they're not on the model

src/QuizAtom.tsx
```diff
@@ -1,7 +1,322 @@
...
+        <input
+            type="radio"
+            id={`answer-${id}`}
+            required
+            checked={isChosen}
+            onClick={() => setChosen(id)}
+            disabled={isAnswered}
+            className={visuallyHidden}
+        ></input>
+        <label
```
gtrufitt, line 276: I think there's some accessibility work required to make the answers tabble-through? 

Testing on https://5f22d2148807980022fc782e-bougkfzsdn.chromatic.com/iframe.html?id=quizatom--default-story

I was seeing a couple of things:

- No focus state as tabbing through the questions, so I was not able to see the current question I was on 
- Because of the 'onClick' as soon as the arrow keys are used to navigate between the radio buttons, the answers are chosen.

I will say though, Frontend doesn't fair much better in this respect (tabbing to an answer - it has a focus state which is good, moving through radio buttons with arrow keys selects my answer):



But it's worth having a think to see if we can improve this in this version.

Maybe you could check the eventType:

```
if(event.type == 'mousedown')
    {
        // do something
        return;
    }
    if(event.type == 'keypress')
    {
        // do something else
        return;
    }`
```

And dealing with things slightly differently 🤔  I'm not 100% sure what this looks like though!

gtrufitt (approved):
This is great! Some dev-exy things and accessibility stuff but it's really nice and succinct and easy to reason with as a whole!

src/QuizAtom.tsx
amyhughes: Thanks Gareth, I wouldn't have thought to check this. @nicl helped me make the quiz useable without a mouse today - thank you.

tabbable

amyhughes: Hmm it looks like I need to work out how to make a gif!

src/QuizAtom.tsx
```diff
@@ -1,7 +1,268 @@
...
+                tabIndex={0}
+                className={cx({
+                    [commonLabelStyle]: true,
+                    [focusLabelStyle]: !isAnswered,
+                    [opaqueGreenBackgroundColour]: isCorrect && isChosen,
+                    [translucentGreenBackgroundColour]:
+                        isCorrect && !isChosen && isAnswered,
+                    [redBackgroundColour]: !isCorrect && isChosen,
+                    [neutralBackgroundColour]: !isCorrect && !isChosen,
+                })}
```
liywjl, line 232: the css here is quite hard to follow. I think `Answer` as a component this might be a good candidate to split out differently.

This is just an idea, but it might be worth splitting out:

- `SelectedAwnser`
- `CorrectAwnser`
- `WrongAwnser`
- `PendingAwnser`

you could then have an emotion css class for each and not need to worry too much about colour combinations.

Also this means we could also avoid having nested ternary as each component would be stateless.

src/QuizAtom.tsx
```diff
@@ -1,7 +1,268 @@
...
+    number: number;
+};
+
+export const Question = ({
+    text,
+    imageUrl,
+    answers,
+    number,
+}: QuestionType & QuestionProps): JSX.Element => {
+    const [chosen, setChosen] = useState<string | undefined>(undefined);
```
liywjl, line 138: I think the variable `chosen` might be better named to `selectedId`, I suspect this is an ID because you use `chosen === answer.id` later down the component?

package.json
```diff
@@ -38,6 +38,7 @@
         "@typescript-eslint/parser": "^3.0.1",
         "babel-loader": "^8.1.0",
         "babel-plugin-emotion": "^10.0.33",
+        "babel-plugin-const-enum": "^1.0.1",
```
liywjl, line 41: When I run locally, I get some yarn lock file updates, it might be worth running `yarn` again and pushing the lock changes

package.json
```diff
@@ -84,8 +85,8 @@
     },
     "jest": {
         "testEnvironment": "jest-environment-jsdom-sixteen",
-        "moduleNameMapper": {
-            "^@guardian/src-foundations/(.*)$": "@guardian/src-foundations/$1/cjs"
-        }
+        "transformIgnorePatterns": [
+            "node_modules/(?!(@guardian/src-foundations|@guardian/types)/)"
+        ]
```
liywjl, line 90: is there a reason for these changes? is it related to `const-enum`?

src/QuizAtom.tsx
```diff
@@ -1,7 +1,268 @@
...
+    display: flex;
+    div {
+        padding-right: 0.5ch;
+    }
+    legend {
+        margin-bottom: 12px;
+    }
+    span {
+        padding-right: 12px;
+    }
```
liywjl, line 85: I personally  would prefer having each style apply to the component directly rather than inherited from a parent component. The issue here is that now we apply the same `padding`/`margin` to all DOM child elements.

src/QuizAtom.tsx
```diff
@@ -1,7 +1,268 @@
...
+`;
+
+const answerStyle = css`
+    display: flex;
+    div {
+        display: flex;
+        flex-direction: column;
+    }
```
liywjl, line 93: same inheritance issue here

src/QuizAtom.tsx
```diff
@@ -1,7 +1,268 @@
...
+`;
+
+const iconStyle = css`
+    height: 26px;
+    svg {
+        fill: ${neutral[100]};
+        height: 26px;
+        width: 26px;
+        view-box: '0 0 26px 26px';
+    }
```
liywjl, line 103: same inheritance issue here

liywjl (changes requested):
Really great work 👍  this is quite a challenging component to implement.

There are some areas I think we could simplify, and also I was unable to run storybook on this branch locally, so I think there are some bits of missing code.

liywjl (approved):

PR #46, opened 2020-08-27
Youtube Atom (née MediaAtom)

nicl:
**UPDATE:** have excluded `Caption` now as the platforms may differ in this and it's a component that is shared more widely too and we'd rather not duplicate it. I've updated the picture accordingly.

## What does this change?

Adds (copied from DC) the Youtube/Media Atom. Note, changes have been made to:

* use Format instead of separate Pillar/Design/Display arguments
* ~~inline icon SVGs (for now)~~ update: use src here where possible

There are a lot of open questions here, which I wanted to get feedback on before continuing:

* ~~do we need *all* of the variants of Caption etc? - I am a bit concerned with the lines of code involved in some of these components, even though they reflect (current) real features/design variations~~
* should we just call this YoutubeAtom? (update: yes)
* how do we handle code reuse for child components, pillar palettes
etc?

## Images

Screenshot 2020-09-07 at 15 23 27


gtrufitt:
> should we just call this YoutubeAtom?

+1

src/MediaAtom/Caption.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { Format, Pillar, Display } from '@guardian/types/Format';
+
+const TriangleIcon = (
+    <svg width="11" height="10" viewBox="0 0 11 10">
+        <path fillRule="evenodd" d="M5.5 0L11 10H0z" />
+    </svg>
+);
+
+type Props = {
+    format: Format;
```
gtrufitt, line 18: Feels weird to have to bring caption into here when it exists in dotcom rendering? I would say that the caption used for all embeds/images/media etc should be the same? +1 on taking a Format!

gtrufitt:
Note on the failing compilation:

`semantic error TS17016: JSX fragment is not supported when using --jsxFactory`

I've seen this elsewhere.

Issue with TS, but can be fixed with babel plugins.
https://github.com/developit/microbundle/pull/623
https://github.com/cision/rover-ui/pull/186

Looks like fixed in TS4.0
https://github.com/microsoft/TypeScript/pull/38720

src/MediaAtom/YoutubeOverlay.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const BottomLeft = ({
+    children,
+}: {
+    children: JSX.Element | JSX.Element[];
+}) => (
+    <div
+        className={css`
+            position: absolute;
+            bottom: 12px;
```
gtrufitt, line 29: Do these magic numbers relate to any particular thing that you can assign to a var/ comment to give them context?

src/MediaAtom/MediaMeta.tsx
```diff
@@ -0,0 +1,139 @@
+import React from 'react';
+import { css } from 'emotion';
+
+import { textSans } from '@guardian/src-foundations/typography';
+
+import { pillarPalette } from './pillarPalette';
+import { Pillar } from '@guardian/types/Format';
+
+const Audio = (
```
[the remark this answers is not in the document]
nicl: I think I just copied them to get something working, but I'll switch to the source ones!

src/YoutubeAtom/YoutubeAtom.tsx
```diff
@@ -0,0 +1,78 @@
...
+                iu: `${adTargeting.adUnit || ''}`,
+                cust_params: encodeURIComponent(
+                    constructQuery(adTargeting.customParams),
+                ),
+            },
+        },
+    };
+};
+
+const constructQuery = (query: { [key: string]: any }): string =>
```
jfsoul, line 35: I suppose there isn't anywhere more sensible for this to live?

jfsoul (approved):
Code looks good. I get the argument for it being called `YoutubeAtom`, but at the same time `Media` is the terminology used elsewhere (i.e. tools) so it might be worth documenting somewhere (where?) that they are intended to be equivalent.

src/MediaAtom/YoutubeOverlay.tsx
nicl: Thanks, I've used `space` here now from Source.

src/YoutubeAtom/YoutubeAtom.tsx
nicl: Yeah, it is pretty generic code. I'm not too concerned about this being duplicated though (if it does end up being useful elsewhere), as it's only a few lines of code. Or else to refactor later as an alternative.

PR #58, opened 2020-09-08
Update README for moving default branch to main

shtukas:
Add instruction to move the default branch from `master` to `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,18 @@ import { TheAtomYouWant } from '@guardian/atoms-rendering';
 <TheAtomYouWant someProp={localData.someProp} />
 ```
 
+## Moving to main
+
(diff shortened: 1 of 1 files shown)


amyhughes (commented):
I'm not sure this belongs in the atoms-rendering readme?

It's not a process which is related to this specific repository. You could arguably put in in the frontend docs, but I'd be hesitant about doing this since the advice will soon be outdated.

shtukas:
@amyhughes Oh sorry. I thought it was the same instructions for this repository as well. I actually followed the same instructions and it worked well. No worries. I can close the PR :)

shtukas:
I have updated the typo in the text. Should be ok now :)

PR #62, opened 2020-09-09
add explicit color style to detail element

webb04:
## What does this change?
Sets an explicit color style on the details element in the expandableAtom component instead of relying on outside page styles.
This would match the `ExplainerAtom` component https://github.com/guardian/atoms-rendering/blob/main/src/ExplainerAtom.tsx#L26

Without setting this color in the atom styles, we have some problems in dark mode on apps (screenshots below).

Users should see no difference.

## Images

apps-rendering dark mode.

| Before | After |
| --- | --- |
|  |  |


nitro-marky (approved):
👍

PR #67, opened 2020-09-10
Remove gu-script from chart-atom

nitro-marky:
Minor change, removes the gu-script tag from chart-atoms as resizing now works without it.


nicl (approved):
⛵

PR #73, opened 2020-09-22
Audio Atom

shtukas:
## What does this change?

Add Audio Atom

# After
Screenshot 2020-09-30 at 11 31 43
Screenshot 2020-09-30 at 11 31 37
Screenshot 2020-09-30 at 11 31 32


shtukas:
+1

jfsoul:
Do we know if there is anything we're missing vs. frontend parity now?

liywjl:
@jfsoul From what I see no, here is an example in frontent:
https://www.theguardian.com/australia-news/2020/jun/19/morning-mail-university-fees-to-skyrocket-abetzs-questionable-claim-swan-urges-restraint?dcr=false

I noticed that we do not have any highlighting happening when we tab through, is this something we would like to add? if so we may need to get some designs to know how to display elements selected.

jfsoul:
@liywjl I think parity with frontend is good for now, although if we can improve accessibility a followup PR would be great.

Speaking of parity, I believe this is the place to look for frontend's JS: https://github.com/guardian/atom-renderer/blob/master/core/src/main/resources/lib/audio.fjs

Apart from that JS the only interesting thing I can see it do is optionally add acast here: https://github.com/guardian/atom-renderer/blob/master/core/src/main/resources/audio/article/index.scala.html#L6

liywjl:
@jfsoul thank you for pulling up these repos 👍 

regarding `/audio.fjs` the only thing I seem to be missing is the Ophan reporting, I am trying to wrap my head around some of the logic but I cannot seem to find where `ophan.record` comes from, could you point me in the direction on where this method is defined?

regarding `/article/index.scala.html#L6` we need to get a new prop `externalAdvertising` this will likely need changes on FE from @shtukas

shtukas:
I can't github approve because I made the early commits, but here is a 👍

PR #98, opened 2020-10-23
Update docs: add examples of adding an atom, and relationship between this lib and other services

amyhughes:
## What does this change?

@nitro-marky and I have been working to make the docs for adding atoms a little clearer, here we add a diagram and an explanation of how this library fits with other services. We also add examples of how to add new atoms to this repo.

Suggestions/comments welcome!


liywjl (approved):
👍

PR #101, opened 2020-10-30
fix window undefined caused by SSR on Youtube block element

liywjl:
## What does this change?
Checks if `window` is defined before using that variable

## Why?
When running this component on DCR, we would get a `window` is undefined. This is because when ServerSideRendering, window does not exist. Therefore we need to make a check that `window` exists before using it.

### DIFF
--- a/src/YoutubeAtom/YoutubeAtom.tsx
+++ b/src/YoutubeAtom/YoutubeAtom.tsx
@@ -81,6 +81,13 @@ export const YoutubeAtom = ({
 }: YoutubeAtomType): JSX.Element => {
     const embedConfig =
         adTargeting && JSON.stringify(buildEmbedConfig(adTargeting));
+
+    // if window is undefined it is because this logic is running on the server side
+    const origin =
+        typeof window !== 'undefined' && window.isStory
+            ? ''
+            : '&origin=https://www.theguardian.com';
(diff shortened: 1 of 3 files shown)


shtukas (approved):

src/YoutubeAtom/YoutubeAtom.tsx
```diff
@@ -81,34 +75,34 @@ export const YoutubeAtom = ({
 }: YoutubeAtomType): JSX.Element => {
     const embedConfig =
         adTargeting && JSON.stringify(buildEmbedConfig(adTargeting));
+
+    // if window is undefined it is because this logic is running on the server side
+    const origin =
+        typeof window !== 'undefined' &&
+        window.location.hostname === 'localhost'
+            ? ''
+            : '&origin=https://www.theguardian.com';
```
JamieB-gu, line 84: It sounds like this specifies the origin of the page in which the embed appears. What happens if this is running on AR instead of DCR, where the origin is `https://mobile.guardianapis.com`?

liywjl: this is a good point, another case would be running on `CODE` where the URL will start within something like `https://m.code.dev-theguardian.com`. I just noticed that the JSON for articles returns `host`, so will look at setting that for the origin instead of this logic 👍

PR #105, opened 2020-11-04
Acast Audio

oliverlloyd:
## What does this change?
We should only serve audio with ads in it if the user has consented to this. This PR adds a check to see if the user has consented and, if they have, it reloads the AudioAtom using a new url, one which points to the Acast service to get ad enabled content.

Note. In the gifs below the 14m track is the one without ads, and the 17m track includes adverts.

## What Next?
Once this PR is merged, we need to

1. Release a new version of `@guardian/atoms-rendering`
2. Add the switch `acastEnabled` to DCR
3. Upgrade the version of `atoms-rendering` used in DCR and add the prop, `shouldUseAcast`, using CMP to check if the user has given consent or not

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -5,7 +5,7 @@ module.exports = {
     addons: ['@storybook/preset-typescript'],
     webpackFinal: async (config, { configType }) => {
         config.module.rules.push({
-            test: /(\.tsx?|\.js?)$/,
+            test: /(.tsx?$/,
             use: [
                 {
                     loader: 'babel-loader',
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -5,7 +5,7 @@ module.exports = {
     addons: ['@storybook/preset-typescript'],
     webpackFinal: async (config, { configType }) => {
         config.module.rules.push({
-            test: /\.tsx?$/,
+            test: /(\.tsx?|\.js?)$/,
             use: [
                 {
                     loader: 'babel-loader',
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -5,7 +5,7 @@ module.exports = {
     addons: ['@storybook/preset-typescript'],
     webpackFinal: async (config, { configType }) => {
         config.module.rules.push({
-            test: /(.tsx?$/,
+            test: /\.tsx?$/,
             use: [
                 {
                     loader: 'babel-loader',
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
         "husky": "^4.2.5",
         "jest": "^26.5.2",
         "jest-environment-jsdom-sixteen": "^1.0.3",
+        "ts-jest": "^24.3.0",
         "microbundle": "^0.12.0",
         "prettier": "^2.0.5",
         "pretty-quick": "^2.0.1",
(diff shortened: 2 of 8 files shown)


.storybook/main.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     addons: ['@storybook/preset-typescript'],
     webpackFinal: async (config, { configType }) => {
         config.module.rules.push({
-            test: /\.tsx?$/,
```
oliverlloyd, line 8: Because we're importing CMP now, Storybook was trying to build using js files that it wasn't configured to transform

src/Answers.tsx
```diff
@@ -66,13 +66,13 @@ export const UnselectedAnswer = ({
             onKeyPress={onKeyPress}
             tabIndex={disabled ? -1 : 0}
             htmlFor={`answer-${id}`}
-            data-answerType={
```
oliverlloyd, line 69: Boyscouting to clear up console warning / errors in jest

src/QuizAtom.tsx
```diff
@@ -108,6 +108,7 @@ export const Question = ({
                                 if (answer.isCorrect) {
                                     return (
                                         <CorrectSelectedAnswer
+                                            key={answer.id}
```
oliverlloyd, line 111: Cleaning up some react warnings on the console

package.json
```diff
@@ -79,16 +81,45 @@
...
     "peerDependencies": {
+        "@guardian/consent-management-platform": "^6.6.0",
         "@guardian/src-foundations": "^2.4.0",
         "@guardian/src-icons": "^2.4.0",
         "emotion": "^10.0.27",
         "react": "^16.13.1",
         "react-dom": "^16.13.1"
     },
     "jest": {
+        "preset": "ts-jest/presets/js-with-babel",
```
oliverlloyd, line 92: We can revisit this but I've copied the jest config from DCR here to fix errors that I was seeing after adding CMP

package.json
```diff
@@ -79,16 +81,45 @@
...
+        "moduleNameMapper": {
+            "^svgs/(.*)$": "<rootDir>/__mocks__/svgMock.tsx",
+            "^(.*)\\.svg$": "<rootDir>/__mocks__/svgMock.tsx",
+            "^@root(.*)$": "<rootDir>$1",
+            "^@frontend(.*)$": "<rootDir>/src$1",
+            "^@guardian/src-foundations/(.*)(?<!cjs)$": "@guardian/src-foundations/$1/cjs"
+        },
         "transformIgnorePatterns": [
-            "node_modules/(?!(@guardian/src-foundations|@guardian/types)/)"
+            "/node_modules/(?!@guardian/).+\\.js$"
```
mxdvl, line 119:
```suggestion
            "/node_modules/(?!@guardian/)"
```

liywjl (approved):
Great work, thanks for tidying up 👍

mxdvl (approved):
This looks good from a CMP perspective.

mxdvl:
Actually, do you know why the blue dot indicating how far along the track a user is jumps about?

It also seems to be slightly off the left edge, when on `00:00`, but that might be for a subsequent PR.

oliverlloyd:
I've updated this PR to move the CMP check out of the atom, replacing it with a `shouldUseAcast` prop instead. It is now the responsibility of the calling component to decide if acast should be used or not.

This adds more work to the consumer but makes the atom simpler.

.storybook/main.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     addons: ['@storybook/preset-typescript'],
     webpackFinal: async (config, { configType }) => {
         config.module.rules.push({
-            test: /(\.tsx?|\.js?)$/,
+            test: /(.tsx?$/,
```
mxdvl:
```suggestion
            test: /(.tsx?)$/,
```

jfsoul (approved):

oliverlloyd:
> Looking into the GIFs, do you know why the blue dot indicating how far along the track a user is jumps about?
> 
> It also seems to be slightly off the left edge, when on `00:00`, but that might be for a subsequent PR.

@mxdvl You're right, this was a bug where the duration was being read as `NaN`, good spot. Fixed now 👍

PR #106, opened 2020-11-06
Hoist YouTube component

oliverlloyd:
Moves the `YouTubeAtom` out of its own folder into the component root.

## Why?
I was motivated to do this work because whilst playing with upgrading Storybook to v6 I noticed some errors that seemed to relate to how this component exported itself. I only briefly looked at the reason for these errors to be honest, before seeing the opportunity to hoist things, so I'm not certain this will fix them but it seemed like a reasonable step to take before returning to the effort of upgrading SB.

Also, I believe the reason a separate folder was being used here is because of `MaintainAspectRatio`, a sort of helper component that was only being used by the YouTube component. Because it is only being used in one place, I've inlined this code to make it easier to find for future delevelopers (no need to switch between files).

Once this file was inlined then there was no different between `YouTubeAtom` and other atoms and the hoist made sense.

## What else?
I removed use of the `format` prop in the jest tests. The `YouTubeAtom` component doesn't have this prop so this looks like unused code.


shtukas (approved):

oliverlloyd:
[Refactor YouTubeAtom](https://trello.com/c/kFD8zhW9/2132-refactor-youtubeatom)

liywjl (approved):
👍

PR #110, opened 2020-11-09
Audio Colours

oliverlloyd:
The little dot on the range slider was not picking up the pillar colour so this adds the styles to do that

### Before
Screenshot 2020-11-09 at 13 02 22

### After
Screenshot 2020-11-09 at 13 01 31

### DIFF
--- a/src/AudioAtom.tsx
+++ b/src/AudioAtom.tsx
@@ -116,6 +116,28 @@ const progressBarInputStyle = (pillar: Pillar) => css`
     :focus {
         ${focusHalo}
     }
+    // Use the pillar to style the colour of the range thumb
+    &::-webkit-slider-thumb {
+        background: ${pillarPalette[pillar][400]};
+        -webkit-appearance: none;
+        width: 14px;
+        height: 14px;
+        border-radius: 50px;
+    }
+    &::-moz-range-thumb {
+        background: ${pillarPalette[pillar][400]};
+        width: 14px;
+        height: 14px;
+        border: none;
+        border-radius: 50px;
+    }
+    &::-ms-thumb {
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Always blue](https://trello.com/c/x1esBaLc/2135-always-blue)

src/AudioAtom.tsx
```diff
@@ -116,6 +116,28 @@ const progressBarInputStyle = (pillar: Pillar) => css`
...
+        border: none;
+        border-radius: 50px;
+    }
+    &::-ms-thumb {
+        background: ${pillarPalette[pillar][400]};
+        width: 14px;
+        height: 14px;
+        border: none;
+        border-radius: 50px;
+    }
```
liywjl, line 140:
```suggestion
    &::-webkit-slider-thumb, &::-moz-range-thumb, &::-ms-thumb {
        background: ${pillarPalette[pillar][400]};
        -webkit-appearance: none;
        width: 14px;
        height: 14px;
        border-radius: 50px;
    }
```

liywjl (approved):
Great catch! The thumb CSS selector is non standard, but that is not something we can control.

src/AudioAtom.tsx
oliverlloyd: In [this css-tricks article](https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/) it says you can't group these selectors like that:

> Note that while we’re repeating code here, that’s necessary as you can’t comma-separate these type of selectors. Browsers will drop the entire selector if it doesn’t understand a part of it.

I haven't tried this though

liywjl: cool! yes then you can ignore

PR #112, opened 2020-11-10
Use Both Node 12 And Node 10 In CI

JamieB-gu:
## Why?

Atoms-rendering is used by DCR (Node 10) and AR (Node 12), so it would be good to make sure it runs on both. These changes borrow heavily from [image-rendering](https://github.com/guardian/image-rendering/blob/cc9b89b96462a8df78d17bae69787597caaa6363/.github/workflows/node.js.yml).

**Note**: The lint warnings are present in `main`.

## Changes

- Added a "matrix" to use both Node 10 and 12 for the "lint", "tsc" and "test" jobs

### DIFF
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -4,10 +4,14 @@ on: [push]
 jobs:
     run:
         runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                node-version: [10.x, 12.x]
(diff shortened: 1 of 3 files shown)


oliverlloyd (approved):
If the checks are ok, then I am too.

Plus I also checked that this built locally and SB still runs okay - all good

PR #126, opened 2020-11-12
Automate the release

oliverlloyd:
## What?
This PR attempts to improve how we release code. It does this in two ways:

1. By adding the Github Action [Release Drafter](https://github.com/marketplace/actions/release-drafter) to aid in creating useful release documentation

Screenshot 2020-11-12 at 11 51 54

2. By adding a `yarn release` script in `package.json` to automate the following steps:

- Building the package
- Bumping package version
- Adding a git tag
- Pushing the tag to `main`
- Publishing to Npm
- Launching a browser on the [releases page](https://github.com/oliverlloyd/publish-sandbox/releases) for the repo

## Usage
The `release` script just runs `yarn version` under the hood and then that uses the `preversion` and `postversion` hooks to create a workflow

Eg.:
`yarn release` - Set version interactively
or
`yarn release -- patch` - 0.0.1
`yarn release -- minor` - 0.1.0
`yarn release -- major` - 1.0.0

Etc. [See docs](https://classic.yarnpkg.com/en/docs/cli/version/)

### Why are we using `HUSKY_SKIP_HOOKS` twice?
`HUSKY_SKIP_HOOKS` tells husky not to run when we push the tags. We have to set this env for _each_  push because (I presume) husky removes it after it sees it.

### Why skip hooks at all?
We skip hooks in the `release` step because we have already run `validate` in the `preversion` stage. Having a separate `validate` command is both more semantic and also lets use have different levels of validation for Husky and for when we publish because we might, say, want more checks for publishing a version than we do for a simple push to the repo.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,0 +1,16 @@
+name: Release Drafter
+
+on:
+    push:
+        # branches to consider in the event; optional, defaults to all
+        branches:
+            - main
+
+jobs:
+    update_release_draft:
+        runs-on: ubuntu-latest
+        steps:
+            # Drafts your next Release notes as Pull Requests are merged into "master"
+            - uses: release-drafter/release-drafter@v5
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
     update_release_draft:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-18.04
         steps:
             # Drafts your next Release notes as Pull Requests are merged into "master"
             - uses: release-drafter/release-drafter@v5
--- a/README.md
+++ b/README.md
@@ -110,7 +110,6 @@ Pre-requisits:
 
 1. Ensure your changes are on main
 2. Ensure you have an [npm account](https://docs.npmjs.com/creating-and-publishing-scoped-public-packages) that is authorised for the npm @guardian organisation
-3. `yarn build`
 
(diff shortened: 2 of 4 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,16 @@
...
+
+on:
+    push:
+        # branches to consider in the event; optional, defaults to all
+        branches:
+            - main
+
+jobs:
+    update_release_draft:
+        runs-on: ubuntu-latest
```
liywjl, line 11: I would try pick a LTE version of `ubuntu` instead of latest

liywjl (changes requested):
I think this would benefit of updating the readMe to reflect how to do a release

README.md
```diff
@@ -104,14 +104,16 @@ Adding a new atom in `atoms-rendering` involves
...
 
-## Publishing to NPM
+## Releasing a new version / Publishing to NPM
 
-Manual publishing steps:
+Pre-requisits:
 
 1. Ensure your changes are on main
 2. Ensure you have an [npm account](https://docs.npmjs.com/creating-and-publishing-scoped-public-packages) that is authorised for the npm @guardian organisation
 3. `yarn build`
```
tjmw, line 113: Could the build happen as a pre-step to releasing so that you don't need to run it manually? I think you'd probably _always_ want to build first?

tjmw (approved):
This looks great and I'll definitely be stealing it at some point in future! 😄

liywjl (approved):
LGTM 👍

README.md
oliverlloyd: Great point. I completely missed that! Done 👍

oliverlloyd:
[Release automation](https://trello.com/c/pd7dMjWq/2147-release-automation)

PR #142, opened 2020-11-17
Add analytics for youtube video

liywjl:
## What does this change?
Adds support for `gaEventEmitter` and `ophanEventEmitter` in YoutubeAtom. In order to do so a lot of logic was lifted from YoutubeOverlay component and we need to now track when a video is played and its progress.

Events that we want to track:
- On video is played
- Reached 25% 
- Reached 50%
- Reached 75%
- Finished video

## How to test
Run `yarn storybook` look at the different youtube stories. You should be able to see when an analytics event emiter is called in the console.

Make sure only one event get fired for each state of the Youtube video


src/YoutubeOverlay.tsx
```diff
@@ -76,102 +74,34 @@ const videoDurationStyles = css`
 export const YoutubeOverlay = ({
```
oliverlloyd, line 74: Maybe we can inline this now that there's no so much logic here?

src/YoutubeAtom.tsx
```diff
@@ -1,10 +1,10 @@
-import React from 'react';
-import { css } from 'emotion';
+import React, { useEffect, useState, useCallback } from 'react';
 
 import { MaintainAspectRatio } from './common/MaintainAspectRatio';
-
 import { YoutubeOverlay } from './YoutubeOverlay';
 
+declare let window: any;
```
oliverlloyd, line 6: This seems strange, isn't `window` already a global?

src/YoutubeAtom.tsx
```diff
@@ -64,8 +66,14 @@ type YoutubeAtomType = {
     title?: string;
     duration?: number; // in seconds
     origin?: string;
+    gaEventEmitter?: (event: VideoEventKey) => void;
+    ophanEventEmitter?: (event: VideoEventKey) => void;
 };
 
+type YoutubeStateChangeEvent = { data: -1 | 0 | 1 | 2 | 3 | 5 };
+
+let progressTracker: NodeJS.Timeout | null;
```
oliverlloyd, line 75: I _think_ this is scoped but it'd be interesting to have some test cases where there are multiple youtube elements on the same page

src/YoutubeAtom.tsx
liywjl: Good point, I think the way we do types in this repo isn't ideal. We should not need to export each type. I think however that would be beyond the scope of this PR. I am happy to put it in the `types.ts` file in the meantime? Currently I havent found any other solution to making TS happy with `window.YT`

src/YoutubeAtom.tsx
liywjl: This is a really good point. I quickly tested it out in a dummy component, looks like it is scoped to that component's rendering so it should be find. But thanks for raising this. If you want to try it out here is the code I used:

`SomeTest.tsx`
```
import React from 'react';

let someRandomNumber;

export const SomeTest = () => {
    someRandomNumber = Math.random();
    return <div>{someRandomNumber}</div>;
};
```

`SomeTest.stories.tsx`
```
import React from 'react';

import { SomeTest } from './SomeTest';

export default {
    title: 'SomeTest',
};

export const SomeTestDefault = () => (
    <>
        <SomeTest />
        <SomeTest />
        <SomeTest />
    </>
);
```

Output:
Screenshot 2020-11-18 at 11 29 49

oliverlloyd: Related, but different issue: It might be good to comment here that the reason this setInterval _isn't_ in a `useEffect` is that it's not setting state. If someone were to come along at a later stage and drop a setMyStateValue hook function call here it [won't work](https://stackoverflow.com/questions/53024496/state-not-updating-when-using-react-state-hook-within-setinterval).

Explanation: https://overreacted.io/making-setinterval-declarative-with-react-hooks/

On the other hand, maybe we can put it in one? We're hydrating things anyway so... But there is a cost to this in terms of the code not running until after the render so we should think this through but as far as I can tell this is not time sensitive code and delaying to post render could even be a perf win

src/YoutubeAtom.tsx
```diff
@@ -64,8 +72,172 @@ type YoutubeAtomType = {
...
+            if (!hasUserLaunchedPlay) {
+                gaEventEmitter && gaEventEmitter('play');
+                ophanEventEmitter && ophanEventEmitter('play');
+            }
+            setHasUserLaunchedPlay(true);
+
+            progressTracker = setInterval(() => {
+                if (!gaEventEmitter)
+                    // eslint-disable-next-line no-console
+                    console.error(`GA function is not available`);
```
gtrufitt, line 123: Do we want to throw this errors on PROD? maybe we want to check `process.ENV` is not prod (well, if you don't swap to have an array of eventEmitters that is!)

src/YoutubeAtom.tsx
```diff
@@ -64,8 +72,172 @@ type YoutubeAtomType = {
...
+        // ended
+        case 0: {
+            gaEventEmitter && gaEventEmitter('end');
+            ophanEventEmitter && ophanEventEmitter('end');
+            progressTracker && clearInterval(progressTracker);
+        }
+    }
 };
 
+const overlayStyles = css`
```
gtrufitt, line 177: This CSS doesn't seem related to the PR? Should it be a separate PR?

src/YoutubeAtom.tsx
```diff
@@ -76,11 +248,117 @@ export const YoutubeAtom = ({
...
+                videoId: `${videoMeta.assetId}`,
+                events: {
+                    onReady: () => setIsPlayerReady(true),
+                    // Issue with setting events on Youtube object (refer to useEffect below)
+                    // https://stackoverflow.com/a/17078152
+                    // onPlayerStateChange,
+                },
+                playerVars: {
+                    mute: 1,
+                    autoplay: 1,
```
gtrufitt, line 277:
```suggestion

                playerVars: {
                    mute: 1,
                    autoplay: 1, // Enabling autoplay because the video loads when a user clicks so when the youtube iframe loads, we want the video to load
```

src/YoutubeAtom.tsx
```diff
@@ -76,11 +248,117 @@ export const YoutubeAtom = ({
...
+            new window.YT.Player(`${videoMeta.assetId}`, {
+                videoId: `${videoMeta.assetId}`,
+                events: {
+                    onReady: () => setIsPlayerReady(true),
+                    // Issue with setting events on Youtube object (refer to useEffect below)
+                    // https://stackoverflow.com/a/17078152
+                    // onPlayerStateChange,
+                },
+                playerVars: {
+                    mute: 1,
```
gtrufitt, line 276: Mute doesn't seem to be a parameter in this list? https://developers.google.com/youtube/player_parameters#Parameters

src/YoutubeAtom.tsx
```diff
@@ -76,11 +248,117 @@ export const YoutubeAtom = ({
...
+    useEffect(() => {
+        // if window is undefined it is because this logic is running on the server side
+        if (typeof window === 'undefined') return;
+
+        if (window.YT) {
+            loadVideo();
+        } else {
+            // If not, load the script asynchronously
+            const tag = document.createElement('script');
+            tag.src = 'https://www.youtube.com/iframe_api';
```
gtrufitt, line 319: We have a `loadScript` util in libs https://github.com/guardian/libs/blob/main/src/loadScript.ts

src/YoutubeAtom.tsx
```diff
@@ -76,11 +248,117 @@ export const YoutubeAtom = ({
...
+            try {
+                player.playVideo();
+            } catch (e) {
+                // eslint-disable-next-line no-console
+                console.error(`Unable to play video due to: ${e}`);
+            }
+        }
+    }, [player, isPlayerReady]);
+
+    const onKeyDownOverlay = useCallback(
```
gtrufitt, line 342: +1 great!

src/YoutubeAtom.tsx
```diff
@@ -95,11 +373,27 @@ export const YoutubeAtom = ({
                 tabIndex={overlayImage ? -1 : 0}
             />
             {overlayImage && (
-                <YoutubeOverlay
-                    image={overlayImage}
-                    duration={duration}
-                    id={videoMeta.assetId}
-                />
+                <div className={hasUserLaunchedPlay ? hideOverlayStyling : ''}>
+                    <img
```
gtrufitt, line 377: Image here has no alt text, do we get passed that? The accessiblity function here is great but it a bit confusing to have a tabble image, can we not make the actual play button functional?

src/YoutubeAtom.tsx
```diff
@@ -64,8 +72,172 @@ type YoutubeAtomType = {
...
+    ophanEventEmitter?: (event: VideoEventKey) => void;
+    getCurrentTime: () => number;
+    getDuration: () => number;
+}) => {
+    /** YouTube API
+            -1 (unstarted)
+            0 (ended)
+            1 (playing)
+            2 (paused)
+            3 (buffering)
```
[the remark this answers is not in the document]
liywjl: I was looking at doing it using `YT.PlayerState`, but the only real easy way of accessing those values is by using `window.YT.PlayerState` which feel unnatural to do in this context. Plus it means redefining them in jest tests as we wont have `YT` object available

src/YoutubeAtom.tsx
```diff
@@ -64,8 +72,172 @@ type YoutubeAtomType = {
...
+// make sure that only 1 event has been sent per video
+let has25PercentEventBeenDispatched = false;
+let has50PercentEventBeenDispatched = false;
+let has75PercentEventBeenDispatched = false;
+export const onPlayerStateChangeAnalytics = ({
+    e,
+    hasUserLaunchedPlay,
+    setHasUserLaunchedPlay,
+    gaEventEmitter,
+    ophanEventEmitter,
```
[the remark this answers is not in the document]
liywjl: good point!!! yes, but then we will just have to assume that all emitters want the same param (which is what the current design assumes)

src/YoutubeAtom.tsx
liywjl: I had to lift up the overlay logic to the parent component. @oliverlloyd then advised to remove the need for the overlay component and to lift it as well. So it not technically part of this PR

src/YoutubeAtom.tsx
liywjl: Originally on frontend we would just set a background to a div for the overlay. We do not have any alt text, so I am not sure how best to proceed.

> can we not make the actual play button functional?
I assume what you mean is to turn the play button into an actual button removing the need for use to have to define tab-able buttons?

src/YoutubeAtom.tsx
liywjl: I tried using this lib but got an error in storybook
```
ERROR in ./node_modules/@guardian/libs/dist/esm/loadScript.js 25:19
Module parse failed: Unexpected token (25:19)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|     };
|     const ref = document.scripts[0];
>     ref.parentNode?.insertBefore(script, ref);
| });
```
I am unsure the cause, I suspect it could be something to do with typescript, webpack and storybook (suspect it doesn't understand `?`)

src/YoutubeAtom.tsx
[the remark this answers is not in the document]
liywjl: I created my own object to map values to states as a work around

src/YoutubeAtom.tsx
oliverlloyd: @amyhughes  and I saw a similar issue when trying to use this function. @sndrs off the top of your head is this a problem our end or in the `/libs` project?

.storybook/main.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     stories: ['../**/*.stories.tsx'],
     webpackFinal: async (config, { configType }) => {
         config.module.rules.push({
-            test: /\.tsx?$/,
+            test: /\.[jt]sx?$/,
```
oliverlloyd, line 7: Can remove?

.storybook/main.js
```diff
@@ -18,6 +18,8 @@ module.exports = {
                                     sourceMap: false,
                                 },
                             ],
+                            '@babel/plugin-proposal-optional-chaining',
+                            '@babel/plugin-proposal-nullish-coalescing-operator',
```
oliverlloyd, line 22: Do we need this?

package.json
```diff
@@ -33,9 +33,12 @@
     },
     "devDependencies": {
         "@babel/core": "^7.12.3",
+        "@babel/plugin-proposal-nullish-coalescing-operator": "^7.12.1",
+        "@babel/plugin-proposal-optional-chaining": "^7.12.7",
         "@babel/plugin-transform-runtime": "^7.12.1",
         "@babel/preset-react": "^7.10.4",
         "@babel/preset-typescript": "^7.12.1",
+        "@guardian/libs": "^1.4.2",
```
oliverlloyd, line 41: Are we using this? I think we started to but decided not to maybe?

src/YoutubeAtom.tsx
oliverlloyd: Actually, ignore me. This is caused by us using an older version of Node

oliverlloyd (approved):
A few comments but looks good!

One question. I can see clicks on cards with videos on smaller screens don't play the video but instead link through to the article. We'll need to handle this at some point but is this something we need to handle now for onwards content?

.storybook/main.js
liywjl: no, looks like it is needed to compile dependencies. Copyied the regex from DCR, it supports JS/TS/JSX/TSX

src/YoutubeAtom.tsx
```diff
@@ -55,32 +62,277 @@ type YoutubeMeta = {
...
+// use booleans make sure that only 1 event has been sent per video
+const eventState: { [key: string]: boolean } = {
+    play: false,
+    end: false,
+    25: false,
+    50: false,
+    75: false,
+};
+
+var youtubeCallback: (e: YT.PlayerEvent) => void | undefined;
```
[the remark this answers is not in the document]
liywjl: I cannot use `const` as we allocate (? not sure that is the right word) it another value later https://github.com/guardian/atoms-rendering/pull/142/files#diff-af82942afd667907a1875907a38a536ddb32fc42bac27581c008387a22952b08R282 . But we can change the `var`, to a `let`. I have tested it and it works.

PR #143, opened 2020-11-19
Use preact render instead of render-to-string from react

gtrufitt:
I was struggling with an error in Dotcom Rendering when trying to upgrade to Webpack 5:

```
  ERROR in ./node_modules/@guardian/atoms-rendering/dist/index.modern.js 1:771-821
  Module not found: Error: Package path ./compat/server is not exported from package /Users/gareth_trufitt/code/dotcom-rendering/node_modules/preact (see exports field in /Users/gareth_trufitt/code/dotcom-rendering/node_modules/preact/package.json)
```

The error came from the use of `import { renderToString } from 'react-dom/server';` in atoms-rendering, but I couldn't for the life of me get Dotcom Rendering to compile it properly, despite the use of it in DCR. I upgraded babel, webpack, preact, preact-render-to-string in both projects, messed around with webpack alias, ensured we were bundling etc but it just wouldn't work.

While this should be functionally identical, the problem will come if anyone uses atoms-rendering in a React project then they'll have to install preact *but* react-dom is [much (much) bigger](https://bundlephobia.com/result?p=react-dom@17.0.1) than [preact](https://bundlephobia.com/result?p=preact@10.5.7) and [preact-render-to-string](https://bundlephobia.com/result?p=preact-render-to-string@5.1.12) - so I am not particularly concerned...

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
         "type": "ssh",
         "url": "https://github.com/guardian/atoms-rendering"
     },
-    "version": "3.0.0-0",
+    "version": "2.4.1",
     "source": "src/index.ts",
     "main": "dist/index.js",
     "module": "dist/index.modern.js",
@@ -101,10 +101,10 @@
         "@guardian/src-foundations": "^2.6.0",
         "@guardian/src-icons": "^2.5.0",
         "emotion": "^10.0.27",
-        "preact": "^10.5.7",
         "preact-render-to-string": "^5.1.12",
         "react": "^16.14.0",
-        "react-dom": "^16.14.0"
+        "react-dom": "^16.14.0",
+        "preact": "^10.5.7"
     },
     "jest": {
         "testEnvironment": "jest-environment-jsdom-sixteen",
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
         "jest": "^26.6.3",
         "jest-environment-jsdom-sixteen": "^1.0.3",
         "microbundle": "^0.12.0",
+        "preact-render-to-string": "^5.1.12",
         "prettier": "^2.1.2",
         "pretty-quick": "^3.1.0",
         "react": "^16.14.0",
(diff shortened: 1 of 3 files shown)


rcrphillips:
[Upgrade to Webpack 5](https://trello.com/c/8aUYN7dw/2122-upgrade-to-webpack-5)

webb04 (approved):
We're not using `preact` in apps-rendering but there's good reasons to switch over to it so happy for this to be merged.

We had a few problems upgrading to Webpack 5 too, I think our tripping point was with Storybook so would be good to see the dcr pull request!

 https://github.com/guardian/apps-rendering/pull/927

gtrufitt:
> We're not using `preact` in apps-rendering but there's good reasons to switch over to it so happy for this to be merged.

Yeah, I think it only adds like 6kb to your build *if you don't have preact already* so it's not hugely painful

package.json
```diff
@@ -4,7 +4,7 @@
         "type": "ssh",
         "url": "https://github.com/guardian/atoms-rendering"
     },
-    "version": "2.4.1",
+    "version": "3.0.0-0",
```
oliverlloyd, line 7: Was this change intentional?

oliverlloyd (approved):
Happy to lean into preact more.

One thing though, do we need to be careful of this: https://github.com/preactjs/preact-render-to-string/issues/18?

gtrufitt:
> One thing though, do we need to be careful of this: [preactjs/preact-render-to-string#18](https://github.com/preactjs/preact-render-to-string/issues/18)?

I don't think so? Html attributes are [case insensitive](https://w3c.github.io/html-reference/documents.html#case-insensitivity) so I think this is probably more of style thing?

package.json
gtrufitt: Nope, I was trying to do a secret alpha release and forgot there was a prerelease script thing.

liywjl (approved):
Good workaround 👍  super strange edge case thought.

PR #146, opened 2020-11-27
No figures, No margins

oliverlloyd:
## What does this change?
Removes use of the `figure` tag as a wrapper and removes any (outer) top and bottom margins. This lets the consumer be responsible for deciding top and bottom margins between elements on the page and also for deciding what root element to wrap each element in.

### Why?
There are obvious styling benefits to being able to set and control spacing from one place, the consumer. Any changes can be implemented there and we avoid the situation where margins compete against other margins.

And whilst semantically, it is correct to wrap an atom in a `figure` element so you could argue that this be done here the advantage to allowing the consumer to do it is it allows things like margin t have an element to be applied to without having to wrap the atom in another div. It's also easier to police the semantic structure of the document when it's decided in one place.


liywjl (approved):
+1 for lifting this

oliverlloyd:
[Delete figure from Atoms Rendering](https://trello.com/c/Z6eVl4SO/2182-delete-figure-from-atoms-rendering)

PR #148, opened 2020-12-01
Handle youtube race condition

liywjl:
## What does this change?

Check `window.YT.Player` is defined and set `onYouTubeIframeAPIReady` early on, not in callback

## How to test
Run on storybook, make sure to refresh the page to check if it doen't load from cache.

Solution has not been tested directly on DCR due to `yarn link` hooks issue

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -296,7 +296,12 @@ export const YoutubeAtom = ({
         if (typeof window === 'undefined') return;
 
         if (window.YT) {
-            loadVideo();
+            // An undocumented ready callback to handle edge cases were window.YT.Player may not be ready for init
+            // https://stackoverflow.com/a/62254596
+            // @ts-ignore
+            window.YT.ready(() => {
+                loadVideo();
+            });
         } else {
             // If not, load the script asynchronously
             loadScript('https://www.youtube.com/iframe_api').then(() => {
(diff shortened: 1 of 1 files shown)


src/YoutubeAtom.tsx
```diff
@@ -296,7 +296,12 @@ export const YoutubeAtom = ({
         if (typeof window === 'undefined') return;
 
         if (window.YT) {
-            loadVideo();
+            // An undocumented ready callback to handle edge cases were window.YT.Player may not be ready for init
+            // https://stackoverflow.com/a/62254596
+            // @ts-ignore
```
oliverlloyd, line 301: My eslint screams at me for using @ts-ignore... Are we able to get around having to do this? But if we do, do we need to ignore the ignore?

liywjl: so the reason we have this is that `YT.ready` is not offically documented (this the youtube types do not define it). 

It is true that there is therefore a risk that using an undocumented API, I could look at using instead `onYouTubeIframeAPIReady`.

I am particularly concenred of a minor race condition edge cases that might arise if `onYouTubeIframeAPIReady` is called by YT but we have not set any function to that variable.

oliverlloyd:
Seems like the problems we're trying to solve using YouTube IFrame Player API here are much harder. [This lib ](https://www.npmjs.com/package/youtube-player)is specifically talks about the two main issues:

> The downsides of using YouTube IFrame Player API are:
>
> Requires to define callbacks in the global scope (window).
Requires to track the state of a player (e.g. you must ensure that video player is "ready" before you can use the API).
youtube-player:
>

And the lib offers solutions in the form of:

> Registers listeners required to establish when YIPA has been loaded.
Does not overwrite global YIPA callback functions.
Queues player API calls until when video player is "ready".

liywjl:
superseded by https://github.com/guardian/atoms-rendering/pull/149

PR #149, opened 2020-12-02
Use youtube-player to simplify Atom

liywjl:
## What does this change?
Use `youtube-player` to simplify interations with youttube embed iframe API. This removes the need to worry about if and when the player has been initialised, and allows use to easily add and remove event emitters to the player without having to implement our own event emiter logic.

## How to test
Launch a video on storybook and check that it plays with and without overlay. You should see logs for each event we track on youtube but looking at the browser's console.

We now depend on the logic from `youtube-player`

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useCallback } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import { css, cx } from 'emotion';
 import YouTubePlayer from 'youtube-player';
 
@@ -112,7 +112,7 @@ export const onPlayerStateChangeAnalytics = ({
     setHasUserLaunchedPlay: (userLaunchEvent: boolean) => void;
     eventEmitters: ((event: VideoEventKey) => void)[];
     player: YoutubePlayerType;
-}) => {
+}): void => {
     switch (e.data) {
         case youtubePlayerState.PLAYING: {
             setHasUserLaunchedPlay(true);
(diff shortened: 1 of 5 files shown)


src/YoutubeAtom.tsx
```diff
@@ -225,6 +231,20 @@ const videoDurationStyles = css`
...
+
+// youtube-player doesn't have a type definition, do we have to create our own based on https://github.com/gajus/youtube-player
+type YoutubePlayerType = {
+    on: (state: string, callback: YoutubeCallback) => YoutubeCallback;
+    off: (callback: YoutubeCallback) => void;
+    loadVideoById: (videoId: string) => void;
+    playVideo: () => void;
+    getCurrentTime: () => number;
+    getDuration: () => number;
+};
```
oliverlloyd, line 246: Can we move this up the page so that it is defined before it's used (more for readability reasons than anything)

src/YoutubeAtom.tsx
```diff
@@ -246,95 +266,50 @@ export const YoutubeAtom = ({
     const [hasUserLaunchedPlay, setHasUserLaunchedPlay] = useState<boolean>(
         false,
     );
+    const [player, setPlayer] = useState<YoutubePlayerType | undefined>();
```
oliverlloyd, line 269: I don't think we need to put the player in state. This component won't rerender so it can just be a scoped variable I guess?

src/YoutubeAtom.tsx
```diff
@@ -246,95 +266,50 @@ export const YoutubeAtom = ({
...
-                onPlayerStateChangeAnalytics({
-                    // ts-ignore is used because onStateChange has different listener to what is actually sent
-                    // @ts-ignore
-                    e,
-                    setHasUserLaunchedPlay,
-                    eventEmitters,
-                    player,
-                });
+        if (videoMeta && !player) {
+            setPlayer(YouTubePlayer(`youtube-video-${videoMeta.assetId}`));
```
oliverlloyd, line 277: As above, I suspect this can just be inline. There _may_ be something around `useLayout` but now I'm just making random stream of consciousness statements

src/YoutubeAtom.tsx
```diff
@@ -246,95 +266,50 @@ export const YoutubeAtom = ({
...
 
-        if (window.YT) {
-            loadVideo();
-        } else {
-            // If not, load the script asynchronously
-            loadScript('https://www.youtube.com/iframe_api').then(() => {
-                // onYouTubeIframeAPIReady will load the video after the script is loaded
-                window.onYouTubeIframeAPIReady = loadVideo;
-            });
+            return () => player.off(listener);
```
oliverlloyd, line 297: Hmm, maybe removing the listener is a good use case for useEffect?

oliverlloyd (commented):
I could well be wrong here but I think we could either:

1. Not use useEffect at all here
or 2. Just use one useEffect for everything and ensure it only runs once (empty deps `[]`) 

happy to chat it through

oliverlloyd:
Note. The argument for _not_ using a `useEffect` hook is that the code inside it will only run _after_ the initial render and then any changes could then cause a second render. If we don't need this behaviour we should avoid it

oliverlloyd (approved):

PR #150, opened 2020-12-02
Fix build by adding optional chaining babel

liywjl:
## What does this change?
Adds babel for optional chaining

## How to test
should fix `yarn build`

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -7,6 +7,7 @@ module.exports = {
     plugins: [
         // this is needed to support `const enum`s - see https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats
         'const-enum',
+        '@babel/plugin-proposal-optional-chaining',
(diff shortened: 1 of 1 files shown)


babel.config.js
```diff
@@ -16,11 +17,7 @@ module.exports = {
     ],
     env: {
         test: {
-            plugins: [
-                '@babel/plugin-transform-runtime',
-                '@babel/plugin-proposal-optional-chaining',
-                '@babel/plugin-proposal-nullish-coalescing-operator',
```
oliverlloyd, line 22: We don't need runtime or coalescing anymore?

oliverlloyd (approved):

babel.config.js
liywjl: we still have runtime, it just is on the same line now so shows up as a diff

PR #151, opened 2020-12-03
fix youtube atom

liywjl:
## What does this change?
Adds some fixes to issues that appeared during DCR integration:

- convert`eventState` keys to string to help check
- add `await` to `player.getCurrentTime` and `player.getDuration`
- reduce interval from 1s to 0.5s (because we are using `async` `await` we risk skipping a second)
- use `let player` outside component, checking if has been defined before init

Potential unforseen race conditions

### DIFF
--- a/src/YoutubeAtom.test.tsx
+++ b/src/YoutubeAtom.test.tsx
@@ -47,10 +47,13 @@ describe('YoutubeAtom', () => {
                 e,
                 setHasUserLaunchedPlay,
                 eventEmitters,
-                // @ts-ignore
                 player: {
                     getCurrentTime,
                     getDuration,
+                    on: () => undefined,
+                    off: () => undefined,
+                    loadVideoById: () => undefined,
+                    playVideo: () => undefined,
                 },
             });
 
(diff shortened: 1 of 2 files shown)


src/YoutubeAtom.test.tsx
```diff
@@ -53,77 +54,11 @@ describe('YoutubeAtom', () => {
                 },
             });
 
-            jest.advanceTimersByTime(3000);
+            jest.advanceTimersByTime(1000);
             expect(eventEmitters[0]).toHaveBeenCalledTimes(1);
             expect(eventEmitters[0]).toHaveBeenCalledWith('play');
         });
 
-        it('should dispatch 25% watched event', () => {
```
liywjl, line 61: these tests are flaky, likely due to the fact they depend on setInterval code that is cleared on every itteration

shtukas (approved):

src/YoutubeAtom.tsx
```diff
@@ -265,29 +273,26 @@ export const YoutubeAtom = ({
...
-    const player = useRef<YoutubePlayerType>();
 
     useEffect(() => {
-        if (!player.current)
-            player.current = YouTubePlayer(
-                `youtube-video-${videoMeta.assetId}`,
-            );
-
-        const listener = player?.current?.on(
+        if (!player && typeof window !== 'undefined') {
```
oliverlloyd, line 278: You don't need to check for window being undefined inside a useEffect

oliverlloyd (commented):
Almost there!

I think though that you might need to manually run linting on this, I remember you saying you
 were having issues and I can see `useRef` is still hanging around which linting would throw errors on

oliverlloyd:
Also, it'd be great to try to capture and bump the work to transfer this testing scope into DCR so that it doesn't get lost at the back of a todo list somewhere

oliverlloyd (approved):

PR #153, opened 2020-12-04
Remove interval logic youtube progress tracker

liywjl:
## What does this change?
- instead of using `setInterval` we recursively use `setTimeout`
The enforces that only 1 interval/script is being called at a time.

- use a previous percentage varible to track difference with current and previous percentage progress 
Previously we would have to call every second (or even more often as you cannot ensure promises would resolve at a specified time) to check the progress of a video. But if we have a previous variable, all we check is the current percentage with the previous and then dispatch events based on the difference. 

- use some shared global state for videos outside of the react component
We have to add and remove `listener` to the player during re-renders. However we want to persiste some data to track thinkgs like: has the play event been disptahced? what was the previous percentage progres? is there a recursive timeout already running?
Setting all of these within the react component would force use to have to re-render. To avoid re-renders we have a mutable object called `sharedYTData` and in order for one YT video not to override the data of another, we added a unique key based on assetId.

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -110,28 +110,24 @@ const intervalProgressTracker = async ({
     const percentPlayed = ((currentTime / duration) * 100) as number;
 
     if (pastProgressPercentage < 25 && 25 < percentPlayed) {
-        pastProgressPercentage = percentPlayed;
-
         eventEmitters.forEach((eventEmitter) =>
             eventEmitter('25' as VideoEventKey),
         );
     }
 
     if (pastProgressPercentage < 50 && 50 < percentPlayed) {
-        pastProgressPercentage = percentPlayed;
-
         eventEmitters.forEach((eventEmitter) =>
             eventEmitter('50' as VideoEventKey),
         );
     }
 
     if (pastProgressPercentage < 75 && 75 < percentPlayed) {
-        pastProgressPercentage = percentPlayed;
-
         eventEmitters.forEach((eventEmitter) =>
             eventEmitter('75' as VideoEventKey),
         );
     }
+
+    pastProgressPercentage = percentPlayed;
 
     // we recursively set set timeout as a way of only having one interval at a time querying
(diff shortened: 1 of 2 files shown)


src/YoutubeAtom.tsx
```diff
@@ -274,26 +318,40 @@ export const YoutubeAtom = ({
...
 
+    // to avoid rerenders we use an object outside of React compont's scope
+    // to make sure data is unique to this video and not overriding other youtube
+    // video's data, we key all the data based on videoMeta.assetId
+    if (!sharedYTData[videoMeta.assetId]) {
+        sharedYTData = {
+            ...sharedYTData,
+            [videoMeta.assetId]: initSharedYTData,
+        };
+    }
```
oliverlloyd, line 329: Why do we need to use global state here instead of component level state? I'm not at the coal face for trying to get this to work so I get that YT might be doing weird stuff, but my instinct is that the the correct solution would be to do this in a more 'reacty' way.

src/YoutubeAtom.tsx
```diff
@@ -90,78 +90,123 @@ export const youtubePlayerState = {
...
-                    !eventState[percentPlayed]
-                ) {
-                    eventEmitters.forEach((eventEmitter) =>
-                        eventEmitter(`${percentPlayed}` as VideoEventKey),
-                    );
-                    eventState[percentPlayed] = true;
-                }
-            }, 500);
+            if (sharedYTData[assetId].progressTrackerTimoutId) {
+                clearTimeout(sharedYTData[assetId].progressTrackerTimoutId);
```
oliverlloyd, line 182: I'm wondering if this is needed? Now we're not using setInterval there shouldn't be a need to clear anything. At the end of each check, we either decide that we're complete, or if not, we set the next timeout check.

If there's a situation where a user might click stop and then a check event fires. I think that in this situation we can check if the video is playing or not and use that to decide whether to set the next check.

liywjl:
superseded by https://github.com/guardian/atoms-rendering/pull/159

PR #159, opened 2020-12-08
Youtube progress tracker interval v2

liywjl:
## What does this change?
- Use `useRef` to store player state
- Remove `onPlayerStateChangeAnalytics`  and instead move event dispatch logic all into useEffect.
- Simplify logic by moving boolean checks of it events dispatched to be within useEffect and not in or out of component state

## How to test
On storybook, play through videos and in console check the events are all fired when they hit specific percentages (and that they are only fired once)

The useEffect is rerendered when eventEmitters changes. This means that events could be retriggered if that object changes. However it is likely that eventEmitters will change at most once, and will likely be done before the user plays the video. So there would be low to no impact.

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -9,7 +9,6 @@ import { SvgPlay } from '@guardian/src-icons';
 
 import { MaintainAspectRatio } from './common/MaintainAspectRatio';
 import { formatTime } from './lib/formatTime';
-import { YoutubeStateChangeEventType } from './types';
 
 type Props = {
     videoMeta: YoutubeMeta;
@@ -28,6 +27,8 @@ declare global {
         onYouTubeIframeAPIReady: unknown;
     }
 }
+
+type YoutubeStateChangeEventType = { data: -1 | 0 | 1 | 2 | 3 | 5 };
 
 type EmbedConfig = {
     adsConfig: {
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -110,28 +110,24 @@ const intervalProgressTracker = async ({
     const percentPlayed = ((currentTime / duration) * 100) as number;
 
     if (pastProgressPercentage < 25 && 25 < percentPlayed) {
-        pastProgressPercentage = percentPlayed;
-
         eventEmitters.forEach((eventEmitter) =>
             eventEmitter('25' as VideoEventKey),
         );
     }
 
     if (pastProgressPercentage < 50 && 50 < percentPlayed) {
-        pastProgressPercentage = percentPlayed;
-
         eventEmitters.forEach((eventEmitter) =>
             eventEmitter('50' as VideoEventKey),
         );
     }
 
     if (pastProgressPercentage < 75 && 75 < percentPlayed) {
-        pastProgressPercentage = percentPlayed;
-
         eventEmitters.forEach((eventEmitter) =>
             eventEmitter('75' as VideoEventKey),
         );
     }
+
+    pastProgressPercentage = percentPlayed;
 
     // we recursively set set timeout as a way of only having one interval at a time querying
     progressTrackerTimoutId = window.setTimeout(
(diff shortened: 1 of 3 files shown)


src/YoutubeAtom.tsx
```diff
@@ -7,9 +7,9 @@ import { palette, space } from '@guardian/src-foundations';
 import { textSans } from '@guardian/src-foundations/typography';
 import { SvgPlay } from '@guardian/src-icons';
 
-import { YoutubeStateChangeEventType } from './types';
 import { MaintainAspectRatio } from './common/MaintainAspectRatio';
 import { formatTime } from './lib/formatTime';
+import { YoutubeStateChangeEventType } from './types';
```
oliverlloyd, line 12: You could just inline this. I don't think it's being used anywhere else?

src/YoutubeAtom.tsx
```diff
@@ -274,25 +197,86 @@ export const YoutubeAtom = ({
...
+
+                    const checkProgress = async () => {
+                        if (!player || !player.current) return null;
+                        const currentTime = await player.current?.getCurrentTime();
+                        const duration = await player.current?.getDuration();
+
+                        if (!duration || !currentTime) return;
+
+                        const percentPlayed = ((currentTime / duration) *
+                            100) as number;
```
oliverlloyd, line 237: Is this `as number` coercion needed?

src/YoutubeAtom.tsx
```diff
@@ -274,25 +197,86 @@ export const YoutubeAtom = ({
...
+
+                        if (!hasSent25Event && 25 < percentPlayed) {
+                            eventEmitters.forEach((eventEmitter) =>
+                                eventEmitter('25' as VideoEventKey),
```
oliverlloyd, line 241: Ditto on the coercion, I suspect we can just call the function with the `25` string?

src/YoutubeAtom.tsx
```diff
@@ -274,25 +197,86 @@ export const YoutubeAtom = ({
...
+                            eventEmitters.forEach((eventEmitter) =>
+                                eventEmitter('75' as VideoEventKey),
+                            );
+                            hasSent75Event = true;
+                        }
+
+                        const currentPlayerState = await player.current?.getPlayerState();
+
+                        if (currentPlayerState !== youtubePlayerState.ENDED) {
+                            // we recursively set set timeout as a way of only having one interval at a time querying
```
oliverlloyd, line 263:
```suggestion
                            // Set a timeout to check progress again in the future
```

src/YoutubeAtom.tsx
```diff
@@ -311,12 +295,15 @@ export const YoutubeAtom = ({
 
             {(overlayImage || posterImage) && (
                 <div
-                    onClick={() => player?.playVideo()}
+                    onClick={() => {
+                        setHasUserLaunchedPlay(true);
```
oliverlloyd, line 299: Does this need to be a state value? Perhaps we can just have it as a standard variable inside the useEffect, along with the other progress trackers?

src/YoutubeAtom.tsx
```diff
@@ -311,12 +295,15 @@ export const YoutubeAtom = ({
...
+                        player.current?.playVideo();
+                    }}
                     onKeyDown={(e) => {
                         const spaceKey = 32;
                         const enterKey = 13;
                         if (e.keyCode === spaceKey || e.keyCode === enterKey)
-                            player?.playVideo();
+                            player.current?.playVideo();
```
oliverlloyd, line 306: We don't set the play state here (as above) so this is a good argument for setting that `hasStartedPlaying` state in response to the listener rather than the click/keypress event

oliverlloyd (approved):
A few comments on some improvements / mini bugs but pretty much there!

src/YoutubeAtom.tsx
liywjl: it use to be used for unit tests, so yes good point ill move it back

src/YoutubeAtom.tsx
liywjl: yes because `currentTime` and `duration` are promises so it could resolve to `NaN`

liywjl: though TS seems happy without now.. cool :) will remove

oliverlloyd (approved):

PR #163, opened 2020-12-22
Quiz button padding

liywjl:
## What does this change?

Add padding to bottom of quiz

## Why
On DCR we need some space between the quiz and the footer content

## Images

Before:
Screenshot 2020-12-22 at 14 20 54

After:
Screenshot 2020-12-22 at 14 20 47


shtukas (approved):

oliverlloyd:
You should get vertical padding from `Figure`. We want to maintain this kind of thing in one place I think so part of my task to bring that component in is to delete this kind of css.

PR #167, opened 2020-12-29
Reset knowledge quiz correctly

liywjl:
## What does this change?
Lift selected answers state logic to parent to enable reset button to clear previous quiz selection


shtukas (approved):

oliverlloyd (approved):
It's a small change but it makes a big difference!

PR #168, opened 2020-12-29
Update youtube props

liywjl:
## What does this change?
- Remove `videoMeta` and instead have each prop defined individually
- Change `overlayImage` to `overrideImage` to better match CAPI

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -13,8 +13,8 @@ import { Picture } from './Picture';
 import { ImageSource, RoleType } from './types';
 
 type Props = {
-    videoMeta: YoutubeMeta;
-    overlayImage?: ImageSource[];
+    assetId: string;
+    overrideImage?: ImageSource[];
     posterImage?: ImageSource[];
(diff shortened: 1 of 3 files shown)


shtukas (approved):

src/YoutubeAtom.tsx
```diff
@@ -73,17 +73,6 @@ const constructQuery = (query: { [key: string]: any }): string =>
...
 
-type YoutubeMeta = {
-    assetId: string;
-    mediaTitle: string;
-    id?: string;
-    channelId?: string;
-    duration?: number;
-    posterSrc?: string;
-    height?: string;
-    width?: string;
```
oliverlloyd, line 84: Does CAPI send any of this data or was it all redundant?

oliverlloyd (approved):
+1 For explicit props. This will make things easier to reason about!

src/YoutubeAtom.tsx
liywjl: its redundant, it was just not deconstructed into individual props

PR #169, opened 2020-12-30
Pillar (with labs)

oliverlloyd:
## What does this change?
Replaces the `Pillar` value from @guardian/types with a locally defined one that is very similar but that includes `labs`

## Why?
Because DCR needs labs supported. There is [an open PR](https://github.com/guardian/types/pull/42) to add labs to the /types repo but I want this to be a discussion and so by creating this workaround we remove the dependency and allow the PR time to be discussed if needed.

### DIFF
--- a/src/AudioAtom.stories.tsx
+++ b/src/AudioAtom.stories.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { css } from 'emotion';
 
-import { Pillar } from '@guardian/types/Format';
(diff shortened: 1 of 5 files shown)


shtukas (approved):

oliverlloyd:
[Atoms rendering uses Pillar](https://trello.com/c/HrtwAdUT/2268-atoms-rendering-uses-pillar)

PR #172, opened 2020-12-31
Cypress selector overlay

liywjl:
## What does this change?
Adds a cypress selector to the youtube overlay element


oliverlloyd (approved):
Nice

PR #177, opened 2021-01-06
Knowledge quiz submit per question

liywjl:
## What does this change?
We add a `submit` and `reset` button per question on the knowledge quiz

## Why?

We use to instantly display correct and incorrect answers once an option was selected, but this was found to not work for keyboard users. Instead we went with the idea of having a submit button at the end of the quiz that would change the state of the quiz to display correct/incorrect answers based on selection.  

But that current implementation means that we display correct answers for questions we may not have answered, plus in terms of UX its not great as you would have to scroll back up and down to see what you got right and wrong.  

This is a compramise that keeps accessibility while maintaining some level of UX by not requiring up and down scrolls.

## How to test
Check out the story, use keyboard and mouse to make sure you can submit both ways

## Images
Screenshot 2021-01-07 at 15 24 43

### DIFF
--- a/src/Answers.tsx
+++ b/src/Answers.tsx
@@ -1,10 +1,9 @@
 import React from 'react';
-import { css, cx } from 'emotion';
+import { css } from 'emotion';
 
 import { SvgCheckmark, SvgCross } from '@guardian/src-icons';
 import { neutral, news, success } from '@guardian/src-foundations/palette';
 import { body, textSans } from '@guardian/src-foundations/typography';
-import { visuallyHidden } from '@guardian/src-foundations/accessibility';
 import { space } from '@guardian/src-foundations';
 
 // We export Radio wrapper styles to override Source Radio buttons to align
@@ -35,16 +34,14 @@ const AnswerWithSVG = ({
     id,
     text,
     supplementText,
-    name,
     isCorrect,
-    answertype,
+    answerType,
 }: {
     id: string;
     text: string;
     supplementText?: string;
-    name: string;
     isCorrect: boolean;
-    answertype: string;
+    answerType: string;
 }): JSX.Element => (
     <div
         className={css`
@@ -75,20 +72,6 @@ const AnswerWithSVG = ({
         >
             {isCorrect ? <SvgCheckmark /> : <SvgCross />}
         </div>
-        <input
-            type="radio"
-            id={id}
-            name={name}
-            data-testid={id}
-            className={cx(
-                css`
-                    ${visuallyHidden}
-                `,
-            )}
-            data-answertype={answertype}
-            required
-            tabIndex={-1}
-        />
         <label
             className={css`
                 color: ${neutral[100]};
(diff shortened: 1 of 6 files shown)


src/Answers.tsx
```diff
@@ -75,20 +72,6 @@ const AnswerWithSVG = ({
         >
             {isCorrect ? <SvgCheckmark /> : <SvgCross />}
         </div>
-        <input
```
liywjl, line 78: There were some issues of loosing focue for keyboard users and we do not need input once a selection has been made.

src/Answers.tsx
```diff
@@ -151,26 +132,14 @@ const AnswerWithoutSVG = ({
             padding-left: ${space[9]}px;
         `}
     >
-        <input
```
liywjl, line 154: ditto keyboard users issue

nitro-marky (approved):

src/KnowledgeQuiz.tsx
```diff
@@ -42,175 +42,143 @@ const fieldsetStyle = css`
...
+                                e.keyCode === spaceKey ||
+                                e.keyCode === enterKey
+                            ) {
+                                // prevent submit
+                                e.preventDefault();
+                                setHasSubmitted(true);
+                            }
+                        }}
+                    >
+                        Submit
```
oliverlloyd, line 149: I'm still not sure the word Submit is correct here. Maybe 'check' or 'reveal'? I think having multiple Submit buttons on one page is maybe a bit semantically, er, well, I guess it just feels off?

oliverlloyd (approved):
Just a few comments. Sorry, a lot of them are on the existing code so not a blocker for this PR but I thought I'd drop them in here anyway - feel free to ignore them

This is perhaps a design question but do we want to think about different styles for the buttons? Now that there are multiple actions should they all be primary? And can they be a bit smaller? Defo one for @HarryFischer - whatever Harry thinks is best!

src/KnowledgeQuiz.tsx
```diff
@@ -42,175 +42,143 @@ const fieldsetStyle = css`
...
-            {imageUrl && (
-                <img
+                        <span
+                            className={css`
+                                padding-right: 12px;
+                            `}
+                        >
+                            {number + '.'}
+                        </span>
+                        {text}
```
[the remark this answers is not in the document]
liywjl: im not a huge fan of using template literals for HTML or using white spacing for spacing

HarryFischer:
@oliverlloyd @liywjl They are all primary? Would be happy if they were a little smaller though - maybe the small version (currently it's medium?)
And I agree that submit feels a little odd when repeated down the page. 'Check' is better IMO

liywjl:
@HarryFischer just updated the text to `Reveal` here is a screenshot with `priority="secondary"` 
is that ok?

Screenshot 2021-01-07 at 14 40 37

PR #178, opened 2021-01-07
Replace Youtube overlay colours with decide palette

nitro-marky:
## What does this change?
The Youtube overlay play buttons and duration now use decidePalette to choose the correct colour based on the pillar.

## Images
Screenshot 2021-01-07 at 14 29 17
Screenshot 2021-01-07 at 14 29 23

### DIFF
--- a/src/YoutubeAtom.stories.tsx
+++ b/src/YoutubeAtom.stories.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { css } from 'emotion';
 
 import { YoutubeAtom } from './YoutubeAtom';
+import { Pillar } from '@guardian/types/Format';
 import { sport } from '@guardian/src-foundations';
 
 export default {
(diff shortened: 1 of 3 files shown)


src/YoutubeAtom.stories.tsx
```diff
@@ -24,6 +25,7 @@ export const DefaultStory = (): JSX.Element => {
                     (e) => console.log(`analytics event ${e} called`),
                 ]}
                 duration={252}
+                pillar={2} // Sport
```
oliverlloyd, line 28: If you `import { Pillar } from '@guardian/types/Format';` into this file then you can replace this with `pillar={Pillar.Sport}` so you won't need the comment and things will carry on working if the underlying numbers change

oliverlloyd (approved):
Looks great!

Just a comment on the stories, we should be using the `Pillar.Sport` pattern there.

src/YoutubeAtom.stories.tsx
nitro-marky: Oh useful, I'll change that

PR #179, opened 2021-01-08
Types @ v1.1

oliverlloyd:
## What?
Bumps @guardian/type to v `1.1.0`

## What else?
Removes usage of optional chaining

## Why?
We want access to the `PhotoEssay` design type in DCR and so we need to update this package in each dependency


oliverlloyd:
[@guardian/types v1.1 for atoms-rendering](https://trello.com/c/fkfU8sZB/2310-guardian-types-v11-for-atoms-rendering)

src/PersonalityQuiz.tsx
```diff
@@ -214,7 +214,7 @@ export const PersonalityQuizAtom = ({
                 <Button
                     type="submit"
                     onClick={onSubmit}
-                    onKeyDown={(e) => {
+                    onKeyDown={(e: any) => {
```
liywjl, line 217:
```suggestion
                    onKeyDown={(e: React.KeyboardEvent<HTMLButtonElement>) => {
```

liywjl (approved):
Great work as always! 👍 I assume a lot of the challenge was figuring out how to get the build to work in storybook?

I am curious what `loader: require.resolve('ts-loader'),` does this try use `ts-loader` from from outside the repo?

src/PersonalityQuiz.tsx
oliverlloyd: Thanks!

src/YoutubeAtom.tsx
```diff
@@ -209,70 +209,81 @@ export const YoutubeAtom = ({
         let hasSent50Event = false;
         let hasSent75Event = false;
 
-        const listener = player.current?.on(
```
oliverlloyd, line 212: @liywjl these diffs look wrong to me - can you check if I'm removing code that I shouldn't be here?

liywjl: you are just removing optional chaining in this case. So its fine in my eyes

oliverlloyd: But if you look at the larger diff, things seen very off, no?

oliverlloyd: Double checked with @liywjl and these diffs are fine

PR #185, opened 2021-01-13
Reverting  changes to use the youtube-nocookie.com in youtube atom

frj:
## What does this change?
This reverts the changes introduced to enable enhanced privacy mode for youtube atoms. This change would have had implications for advertising revenue for guardian videos hosted on youtube. 

The other youtube components on the site will continue to use the youtube-nocookie.com

This is consistent with the current implementation for details see:

https://github.com/guardian/frontend/pull/21747

There should be no drop in revenue when live blogs and video pages are served by DCR


oliverlloyd (approved):
👍

PR #187, opened 2021-01-15
Share quiz buttons

liywjl:
## What does this change?
Add share buttons for Knowledge Quiz

## How to test
Go through Personality Quiz (on storybook), on submit you should see the results. You should also see the differnt types of submit buttons present

## Images
Screenshot 2021-01-18 at 13 28 05


oliverlloyd (commented):
Perhaps a simple approach such as:

```
showFacebook && <FacebookIcon url={facebookUrl} />
showTwitter && <TwitterIcon url={facebookUrl} />
```

Might be easier to reason about?

oliverlloyd:
Also, we already have these icon buttons on DCR (and I'd guess also in other places too) so perhaps now is the time to hoist them into Source?

liywjl:
> Also, we already have these icon buttons on DCR (and I'd guess also in other places too) so perhaps now is the time to hoist them into Source?

Yeah that sounds like a good idea!! I will raise an issue on source for these. I think its easy to see the SVGs becoming part of  source, do you think the share icon component itself should probably be hoisted completely in Source?

oliverlloyd (changes requested):
I just cheked source and actually we already have these [icons](https://theguardian.design/2a1e5182b/p/3892a2-icons/i/13404913) and [buttons](https://theguardian.design/2a1e5182b/p/435225-button/t/245449) there so we should be using them

liywjl:
> Perhaps a simple approach such as:
> 
> ```
> showFacebook && <FacebookIcon url={facebookUrl} />
> showTwitter && <TwitterIcon url={facebookUrl} />
> ```
> 
> Might be easier to reason about?

This components is a copy from DCR. I think it would be a great candidate to lift to some global Source based component. I have tried simplifying the logic from what was present on DCR

oliverlloyd (commented):
I don't think we should be building our own icons (plus having to make the associated webpack changes and declaring global types to support them) in this repo. We already have social icons in Source, if there are some missing then adding them to Source should be a dependency for this PR.

In the long run it would probably be less effort to add them to Source than update this PR vs. getting atoms working with svgs, adding our own custom ones, _then_ adding them to Source and then have to refactor atoms again to use them (not forgetting to update webpack/typescript to remove the svg config!)

src/SharingIcons.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { css } from 'emotion';
+
+import { from } from '@guardian/src-foundations/mq';
+import { LinkButton } from '@guardian/src-button';
+
+import { SvgTwitter, SvgFacebook, SvgEnvelope } from '@guardian/src-icons';
+import { WhatsappIcon } from './icons/whatsapp';
+import { MessengerIcon } from './icons/messenger';
+import { LinkedInIcon } from './icons/linked-in';
+import { PinterestIcon } from './icons/pinterest';
```
oliverlloyd, line 11: These should all come from Source

src/SharingIcons.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export const SharingIcons = ({
+    sharingUrls,
+    displayIcons,
+}: {
+    sharingUrls: {
+        [K in SharePlatformType]?: {
+            url: string;
+            userMessage: string;
+        };
```
oliverlloyd, line 49: This type is already defined in types as `SharingUrlsType`

oliverlloyd (commented):
I still think that the logic here could be simplified by just in lining each icon. I'm noot sure that we need to reduce and map over arrays, esp. when we _know_ the list of items.

Essentially, for each possible icon you only need one line:

```
showFacebook && <FacebookIcon url={facebookUrl} />
```

You can define `showFacebook` with `mainList.includes('facebook') && showForMobile` or similar but the point is I'm not sure you need the `SharingIcons` abstraction.

The advantrage to reduce and map is if we add a new social sharing icon then that will be easier. But I suspect that will be rare and even then adding it inline won't be that hard - in might even be easier if you don't know the code

oliverlloyd:
> do you think the share icon component itself should probably be hoisted completely in Source?

I don't have enough context on what is needed for these icons but my guess is no, we just want to add the icons. We already have src-button so joining them together is pretty simple.

liywjl:
share buttons are being added to source https://github.com/guardian/source/pull/693

shtukas (approved):

src/SharingIcons.tsx
```diff
@@ -1,14 +1,18 @@
...
-import { PinterestIcon } from './icons/pinterest';
+import {
+    SvgTwitter,
+    SvgFacebook,
+    SvgEnvelope,
+    SvgWhatsApp,
+    SvgMessenger,
+    SvgLinkedIn,
+    SvgPinterest,
+} from '@guardian/src-icons';
```
oliverlloyd, line 15: 😻

oliverlloyd (approved):
I've changed my mind. Let's not bother using these Source icons, you can keep using the local ones instead 😉

PR #192, opened 2021-01-25
Update node v14.15

liywjl:
## What does this change?
Update Node to most recent LTS (v14.15)

## Why?
Node 10 EOL is upcoming

### Node release schedule
https://nodejs.org/en/about/releases/


nitro-marky (approved):
Looks good

PR #193, opened 2021-01-26
Add Data Attributes for Teleporter

nitro-marky:
## What does this change?

Add data attributes that will allow Telelporter for atoms to pick these up once Teleporter has gone through the Chrome validation.


liywjl (approved):
👍

PR #194, opened 2021-01-27
Update to @emotion/core

liywjl:
## What does this change?
Update the use of emotion to be more in line with how you write emotion 10.

## Why?
We have been using Emotion 10 but we write emotion code as if we are using Emotion 9 (this was allowed to help make the transition simple). But now as we are aiming to go to Emotion 11, we should first migrate to using Emotion 10 syntax.


babel.config.js
```diff
@@ -1,20 +1,18 @@
 module.exports = {
     presets: [
-        '@babel/preset-env',
         '@babel/preset-typescript',
-        '@babel/preset-react',
-    ],
-    plugins: [
-        // this is needed to support `const enum`s - see https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats
-        'const-enum',
```
oliverlloyd, line 9: This isn't related to emotion. I think this is just being removed because it's unused?

babel.config.js
```diff
@@ -1,20 +1,18 @@
...
-        '@babel/plugin-proposal-optional-chaining',
```
oliverlloyd, line 10: Also removed because it's not required?

package.json
```diff
@@ -53,12 +55,11 @@
         "@typescript-eslint/parser": "^3.10.1",
         "babel-loader": "^8.2.2",
         "babel-plugin-const-enum": "^1.0.1",
-        "babel-plugin-emotion": "^10.0.33",
         "chromatic": "^5.4.0",
-        "emotion": "^10.0.27",
         "eslint": "^7.13.0",
         "eslint-config-airbnb-base": "^14.1.0",
         "eslint-config-prettier": "^6.15.0",
+        "eslint-plugin-emotion": "^10.0.0",
```
oliverlloyd, line 62: Are we using this?

src/InteractiveAtom.tsx
```diff
@@ -16,16 +16,12 @@ export const InteractiveAtom = ({
     id,
     html,
     js,
-    css,
+    atomCss,
```
oliverlloyd, line 19: I wonder if it might be better to rename all these props in the same pattern?

src/InteractiveAtom.tsx
```diff
@@ -16,16 +16,12 @@ export const InteractiveAtom = ({
...
-        className={containerStyles}
-        data-atom-id={id}
-        data-atom-type="interactive"
-    >
+    <div css={containerStyles} data-atom-id={id} data-atom-type="interactive">
         <iframe
-            className={fullWidthStyles}
-            srcDoc={unifyPageContent({ js, css, html })}
+            css={fullWidthStyles}
+            srcDoc={unifyPageContent({ js, css: atomCss, html })}
```
oliverlloyd, line 24: Do we want to carry through the prop renaming to here as well?

src/InteractiveAtom.tsx
oliverlloyd: Maybe `elementCss`? Not really sure though 🤷

oliverlloyd (approved):
Exciting work!

Just a few comments but nothing major

nitro-marky (approved):

PR #197, opened 2021-01-28
Fix timeline events bug on firefox

buck06191:
## What does this change?

Moves atoms-rendering to use the unixDate information passed through to fix a bug with Firefox not being able to parse dates in the format `month-year` e.g. "September 2011".

### Related PRs
- https://github.com/guardian/frontend/pull/23504 (required to be live before this merges)
- https://github.com/guardian/dotcom-rendering/pull/2427

## How to test

Run all PRs locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
         "type": "ssh",
         "url": "https://github.com/guardian/atoms-rendering"
     },
-    "version": "11.0.0-0",
+    "version": "10.0.2",
     "source": "src/index.ts",
     "main": "dist/index.js",
     "module": "dist/index.modern.js",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
         "type": "ssh",
         "url": "https://github.com/guardian/atoms-rendering"
     },
-    "version": "10.0.2",
+    "version": "11.0.0-0",
     "source": "src/index.ts",
     "main": "dist/index.js",
     "module": "dist/index.modern.js",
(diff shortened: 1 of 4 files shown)


nitro-marky (approved):

package.json
```diff
@@ -4,7 +4,7 @@
         "type": "ssh",
         "url": "https://github.com/guardian/atoms-rendering"
     },
-    "version": "10.0.2",
+    "version": "11.0.0-0",
```
gtrufitt, line 7: I think just 11.0.0

buck06191: That's come from the pre-major release using yarn release. I assume that if I merge to main and then release it as a major release that should fix it?

src/fixtures/timelineAtom.tsx
```diff
@@ -6,78 +6,91 @@ const NewsEvents = [
         date: '1999',
         body:
```
gtrufitt, line 7: Probably worth adding some fixtures with a `toDate` and `toUnixDate` fields so that we capture them in chromatic

gtrufitt (approved):
Would be good to have toUnixDate in a story but otherwise

package.json
gtrufitt: You'll then have to PR the change again though, so your best bet is to set it here and then ensure you release as 11.0.0 (I think that's correct @oliverlloyd ?)

src/fixtures/timelineAtom.tsx
buck06191: The Julian Assange one should cover this already, but it looks like it only has those field on the first event. I wonder if something has changed if that was meant to be on all of them previously?

package.json
oliverlloyd: So actually you never want to have a PR with changes to `version` in it.

If you're releasing a new version in the normal course of things (non pre release) then it's done directly on `main` so there's no PR

If you're releasing a pre release, like here, then that _can_ be done on the branch (using `--any-branch`) but you don't need to carry that `version` diff into main. You can delete this diff and then when you do the main release on `main` with `yarn release` it will bump it as required.

gtrufitt: Right, perfect, thanks!

PR #199, opened 2021-01-29
Correct the `InteractiveAtom` type

buck06191:
## What does this change?

Updates the prop types of `AtomInteractive` following a recent update that made some optional props compulsory.


liywjl (approved):
🚀

PR #200, opened 2021-01-29
Italics now work in Explainers and a story has been added

nitro-marky:
## What does this change?

Before all text was inheriting the page style, which overrode any italic styles, especially in the poetry. This should now be fixed

## Images
Before:
Screenshot 2021-01-29 at 15 15 59

After:
Screenshot 2021-01-29 at 15 06 03

### DIFF
--- a/src/ExplainerAtom.tsx
+++ b/src/ExplainerAtom.tsx
@@ -30,6 +30,9 @@ const Container = ({
                     margin-top: ${space[3]}px;
                     margin-bottom: ${space[3]}px;
                 }
+                i {
+                    font-style: italic;
+                }
             `}
         >
             {children}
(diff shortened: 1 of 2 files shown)


src/ExplainerAtom.tsx
```diff
@@ -48,16 +51,22 @@ const Title = ({ title }: { title: string }) => (
...
-        css={css`
-            ${textSans.small({ fontWeight: 'light', lineHeight: 'tight' })}
-        `}
-        dangerouslySetInnerHTML={{
-            __html: html,
-        }}
-    />
-);
+const Body = ({ html }: { html: string }) => {
+    console.log(html);
```
gtrufitt, line 55:
```suggestion
```

gtrufitt: Actually, you probably want to revert this back to the `=> ()` rather than `return`?

nitro-marky: Yeah silly mistake, just fixed

gtrufitt (approved):

PR #203, opened 2021-02-03
Manage typescript version in package.json

liywjl:
## What does this change?
- Explicitly set typescript version in `devDependencies` and `peerDependencies`
- VSC fetch typescript version for `node_modules`

## Why?
Using JSX instead of React throws TS issues for older versions of TS. To avoid these issues we should remove the ambiguity of what TS version is running in the repo

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
         "pretty-quick": "^3.1.0",
         "react": "^16.14.0",
         "react-dom": "^16.14.0",
-        "ts-jest": "^24.3.0"
+        "ts-jest": "^24.3.0",
+        "typescript": "^4.1.3"
     },
     "eslintConfig": {
         "root": true,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -114,7 +115,8 @@
         "preact": "^10.5.7",
         "preact-render-to-string": "^5.1.12",
         "react": "^16.14.0",
-        "react-dom": "^16.14.0"
+        "react-dom": "^16.14.0",
+        "typescript": "^4.1.3"
```
buck06191, line 119: We add it to `devDependencies` and `peerDependencies` here - does that need to happen or can it be defined once? My `npm`/`yarn` knowledge isn't enough to know how that works.

buck06191 (approved):
LGTM but with a question around the duplication

package.json
liywjl: interesting point, I assume we do not technically need is in `peerDependencies`, but I think that enforcing TS version could be a nice thing to do especailly as v3 TS doesnt support JSX.

buck06191: That makes sense then. We're saying that running this locally you need this typescript version, but if you use this package we also expect you have to have this version as well. That way we should be enforcing it between packages too. 

Sounds like a plan to me

PR #204, opened 2021-02-03
Update react 17

liywjl:
## What does this change?
Update to React 17 and remove `React` from import scope

## Why?
Emotion 10 uses JSX pragma, and `React` 17 favors JSX over JS. To make Emotion upgrade easier, we should first update `React`

## Changes
- Update React, react types
- Remove React as an import
- Update `@babel/preset-react` options to use `runtime: 'automatic',` and `importSource: '@emotion/core',`
- Add `eslint-plugin-react-hooks`

## Link to supporting Trello card
https://trello.com/c/Wlpg47zl/2410-update-to-react-17

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -7,15 +7,22 @@ module.exports = {
             options: {
                 presets: [
                     '@babel/preset-typescript',
-                    '@emotion/babel-preset-css-prop',
+                    [
+                        '@babel/preset-react',
+                        {
+                            runtime: 'automatic',
+                            importSource: '@emotion/core',
+                        },
+                    ],
                     [
                         '@babel/preset-env',
                         {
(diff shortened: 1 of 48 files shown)


buck06191:
> Emotion 10 uses JSX pragma, and React 17 favors JSX over JS. To make Emotion upgrade easier, we should first update React

I don't think the pragma appears in this PR, so we might want to change this in the description to just be "[uses Emotion's `jsx` function instead of `React.createElement` and the `css` prop](https://emotion.sh/docs/css-prop#get-started)"

This should (hopefully) make it easier for future developers to trace the problem solving logic here

liywjl:
superseded https://github.com/guardian/atoms-rendering/pull/215

PR #216, opened 2021-02-24
Migrate to rollup

liywjl:
## What does this change?
Moves away from `microbundle` and to using `rollup`

## Why?
- With the introduction of JSX, it is not clear how to configure using `microbundle`
- `microbundle` commnuity has less activity than `rollup`
- `discussion-rendering` uses `rollup`, this helps us keep things inline

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -11,14 +11,14 @@ import { SvgPlay } from '@guardian/src-icons';
 import { MaintainAspectRatio } from './common/MaintainAspectRatio';
 import { formatTime } from './lib/formatTime';
 import { Picture } from './Picture';
-import { ImageSource, RoleType } from './types';
+import { ImageSource, RoleType, AdTargetingType } from './types';
 import { Theme } from '@guardian/types';
 
 type Props = {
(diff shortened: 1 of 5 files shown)


src/YoutubeAtom.tsx
```diff
@@ -46,13 +46,7 @@ type EmbedConfig = {
     };
 };
 
-interface AdTargeting {
-    adUnit: string;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    customParams: { [key: string]: any };
-}
```
liywjl, line 53: this is already defined in `./types`

src/types.ts
```diff
@@ -1,6 +1,6 @@
 import { Theme } from '@guardian/types';
 
-export type AdTargeting = {
+export type AdTargetingType = {
```
liywjl, line 3: we should Type as a suffix

src/YoutubeAtom.tsx
liywjl: rollup was complaining of similar type definitions

shtukas (approved):

PR #228, opened 2021-03-16
Replace rollup with tsc

faresite:
# Why

@JamieB-gu and I paired on this.

With `v12.1.0` we have fixed module import problems, however atom-rendering seems to be baking in emotion code in the bundle. for example bundling the emotion `jsx` function instead of importing it from emotion's `jsx-runtime` module. The emotion built in apps-rendering didn't recognise it as part of emotion and couldn't figure out how to optimise it. This prevented us from hoisting the styles into a single style tag. Therefore emotion instead peppered the DOM with style tags which prevented us from hashing them server side for our CSP.

The idea to remove rollup and use tsc instead was inspired by https://github.com/guardian/source/pull/737 . We have noticed that we didn't have any issues with source's code on emotion 11 and decided to follow their path. @sndrs 

## What does this change?
- Removes rollup
- Removes babel
- Uses tsc instead

## How to test
We've tested with apps-rendering.


rollup.config.js
```diff
@@ -1,41 +0,0 @@
...
-            format: 'cjs',
-        },
-        {
-            file: 'dist/index.esm.js',
-            format: 'esm',
-        },
-    ],
-    external: [
-        // Ignore all dependencies and PeerDependencies in dist
-        ...Object.keys(pkg.dependencies || {}),
```
sndrs, line 24: i think if you just want to fix the bundling you can get rollup to exclude emotion's jsx stuff from the build by explicitly adding `@emotion/react/jsx-runtime` to rollup's externals.

this line will ensure `@emotion/react` isn't bundled but it won't match on submodes/paths

sndrs:
it looks like before you were creating an `index.js` and `index.esm.js` but the package is referencing these:

```js
    "source": "src/index.ts",
    "main": "dist/index.js",
    "module": "dist/index.modern.js", // was this and umd version ever created?
    "unpkg": "dist/index.umd.js",
```

[this whole area is unclear](https://jasonformat.com/enabling-modern-js-on-npm/), but it might be good to keep providing a commonjs build to `main` and an `index.esm.js` to `module`. if you are also then explicit about where your types live, it should (hopefully) minimise the chances of ts confusing files (for example, [that's how source does it](https://github.com/guardian/source/blob/main/src/core/foundations/package.json#L9-L11))

gtrufitt:
> ```js
>   "main": "dist/index.js",
>     "module": "dist/index.modern.js", // was this and umd version ever created?
>     "unpkg": "dist/index.umd.js",
> ```

Yeah, we were using https://github.com/developit/microbundle previously, which is, I believe what was being referenced here, but that has been lost in translation. I'm suprised to see that the index files disappear altogether in this PR, where's the entry point?

rollup.config.js
faresite: Thanks @sndrs for your comments. Apart from getting rollup to exclude emotion's jsx, there's the benefit of getting rid of extra dependencies (rollup + babel), inspired by source that seems to do the job! Would it not be best to simplify our *-rendering projects by only relying on the TS compiler if there's no need for an actual bundler and let the the host app bundlers take care of themselves? Not to mention there's a lot less config code one needs to understand/maintain.

faresite: I guess what I don't know is, what is the benefit of adding a bundler here. Maybe there is a reason that I'm not aware of?

faresite:
> it looks like before you were creating an `index.js` and `index.esm.js` but the package is referencing these:
> 
> ```js
>     "source": "src/index.ts",
>     "main": "dist/index.js",
>     "module": "dist/index.modern.js", // was this and umd version ever created?
>     "unpkg": "dist/index.umd.js",
> ```
> 
> [this whole area is unclear](https://jasonformat.com/enabling-modern-js-on-npm/), but it might be good to keep providing a commonjs build to `main` and an `index.esm.js` to `module`. if you are also then explicit about where your types live, it should (hopefully) minimise the chances of ts confusing files (for example, [that's how source does it](https://github.com/guardian/source/blob/main/src/core/foundations/package.json?rgh-link-date=2021-03-16T13%3A27%3A06Z#L9-L11))


👍  thanks for noticing the missing `"types"`.
Is there a benefit to providing commonjs. I'm assuming both `dotcom-rendering` and `apps-rendering` bundle with esm?

faresite:
> > ```js
> >   "main": "dist/index.js",
> >     "module": "dist/index.modern.js", // was this and umd version ever created?
> >     "unpkg": "dist/index.umd.js",
> > ```
> 
> Yeah, we were using [developit/microbundle](https://github.com/developit/microbundle) previously, which is, I believe what was being referenced here, but that has been lost in translation. I'm suprised to see that the index files disappear altogether in this PR, where's the entry point?

@gtrufitt If I understand that correctly then the entry file is in the `tsonfig.json`?
```
"files": ["src/index.ts", "src/types.ts", "src/content.d.ts"],
```

rollup.config.js
sndrs: yeah i completely agree, i think TS compiler is prob the best option for internal packages, i only wanted to mention it in case you'd have stuck with rollup if you could

JamieB-gu:
Good points on the index files @gtrufitt and @sndrs. I have and existing PR (#226) open to deal with dropping `unpkg` and fixing the pointer to `index.esm.js`.

> Is there a benefit to providing commonjs. I'm assuming both `dotcom-rendering` and `apps-rendering` bundle with esm?

This seems reasonable to me; are those are likely to be the only consumers for the foreseeable future?

I think we have two options here:

1. If we decide that ES modules are all we need then we drop the `module` field.
2. If we decide that we also need commonjs then we add a separate script to generate that with `tsc` (passing the `module` flag via the CLI), and tweak the current script to output `index.esm.js`.

sndrs:
> Is there a benefit to providing commonjs

yeah this is a big question, and currently we just [recommend publishing internally es modules](https://github.com/guardian/recommendations/blob/master/npm-packages.md#package-content).

i have heard an argument (I think from @nicl – apologies if i've got that wrong) that npm packages should just work with node, and that's going to mean commonjs modules in `main` for a while. i can see the logic in it.

then you can provide es modules to bundlers etc via the `module` field.

the question is: will our packages be used in node this way? to which the answer is maybe "maybe?".

if you want to go for it you could add by adding an extra build task, something like:

```json
    "main": "dist/commonjs/index.js",
    "module": "dist/esm/index.js", 
    ...
        "build": "tsc --outDir dist/esm",
        "build:node": "tsc --outDir dist/commonjs --module CommonJS",
```

tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
     "compilerOptions": {
-        "target": "esnext",
-        "module": "esnext",
+        "target": "ES2020",
+        "module": "ES2020",
         "lib": ["es2017", "es7", "es6", "dom", "DOM.Iterable"],
```
sndrs, line 5: i think you might want to remove these, as far as i understand it these amount to polyfills, which it's best not to include

JamieB-gu: Type polyfills rather than JS polyfills?

package.json
```diff
@@ -6,23 +6,24 @@
...
     },
     "files": [
         "dist/**/*"
     ],
     "scripts": {
-        "build": "rollup --config && yarn createTsDec",
-        "createTsDec": "tsc --emitDeclarationOnly true --declaration true --declarationDir dist/ --noEmit false",
+        "build": "yarn build:esm && yarn build:commonjs",
+        "build:esm": "tsc",
+        "build:commonjs": "tsc --outDir dist/commonjs --module CommonJS",
```
sndrs, line 21: you might want to target ES2018 here, to be sure it will run happily in node 10 and above
```suggestion
        "build:commonjs": "tsc --outDir dist/commonjs --module CommonJS --target ES2018",
```

JamieB-gu: Does that clash with [our recommendations](https://github.com/guardian/recommendations/blob/master/npm-packages.md#package-content)?

tsconfig.json
sndrs: i think it makes both available? https://www.typescriptlang.org/tsconfig#lib

> TypeScript also includes APIs for newer JS features matching the target you specify; for example the definition for Map is available if target is ES6 or newer.

sndrs: ah no ignore this comment, the stuff it inserts by tartgetting older versions of JS is not related to lib stuff.

does this mean we _must_ include it? or does ts not include what it need by default?

package.json
sndrs: it [won't](https://github.com/guardian/recommendations/pull/55/files#diff-105c2556445e6cf4212095164b2c536c0f3197ec34872e7368dde59a6b6fde36R27)

sndrs (approved):
great!

tsconfig.json
JamieB-gu: I'm a little bit puzzled by the docs tbh. I think they're saying that `lib` will allow you to explicitly set what type definitions are available for different language features/environments. What I'm not clear on is whether this is just a concern for the package author (i.e. these are just used at compile time), or whether these definitions also get included in the declaration files that you generate alongside your package for consumers to use.

PR #248, opened 2021-05-27
Add InteractiveLayoutAtom

OllysCoding:
## What does this change?

Adds an interactive layout atom which renders interactive atoms within the main document instead of in an iframe.

At this time storybook is not included as we had trouble getting it to render - this will be done before interactives go live.

## How to test

N/A

This should help more interactives start to render correctly in DCR

## Images

N/A

## Accessibility

N/A

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)


nicl:
(Have updated to a working storybook example. Though note, it is hard to find an example that works perfectly here as typicaly these atom types rely on markup on the actual web pages that doesn't exist here.)

nicl (approved):
👍

OllysCoding:
> (Have updated to a working storybook example. Though note, it is hard to find an example that works perfectly here as typicaly these atom types rely on markup on the actual web pages that doesn't exist here.)

Looks good! Nice one :)

PR #256, opened 2021-06-07
Correctly sets YouTube atom data attributes

SHession:
## What does this change?

This PR aims to fix a user reported issue with the Teleporter `Copy Embed Link` behaviour for YouTube atoms. The way the current data attributes are configured means the outputted link is incorrect. The `data-atom-id` should be set to the id of the media atom and the `data-atom-type` should be set to "media".

This change is designed to account for the issues a change to the interface might cause by making the new property optional. The upgrade is also only significant for DCR, as this is what Teleporter uses.

## How to test

I have attempted to test this using npm link but I found the video player didn't work with or without my change. It does however work when the package isn't linked. I would be interested in pairin g on testing this if anyone knows where I'm going wrong. 

Users can press the `Copy Embed Link` button and be confident it will work as expected.

Any change to the interface can be frustrating to implement in each of the consumers. This change attempts to mitigate problems by making the necessary parameter optional.

## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [x] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -28,6 +28,7 @@ type Props = {
     origin?: string;
     eventEmitters: ((event: VideoEventKey) => void)[];
     pillar: Theme;
+    atomId?: string;
 };
 declare global {
     interface Window {
@@ -181,6 +182,7 @@ export const YoutubeAtom = ({
     origin,
     eventEmitters,
     pillar,
+    atomId,
 }: Props): JSX.Element => {
     const embedConfig =
         adTargeting && JSON.stringify(buildEmbedConfig(adTargeting));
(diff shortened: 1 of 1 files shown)


shtukas (approved):

joecowton1:
@SHession do you know if this is still required?

PR #270, opened 2021-07-21
`react-dom` Over `preact-render-to-string`

JamieB-gu:
## What does this change?

This switches from using `preact-render-to-string` to `react-dom/server` for the `renderToString` function.

We use a mix of React and Preact across the department, and [common](https://github.com/guardian/dotcom-rendering/blob/14779a173d9f19af8dbab456749f1421d5a372fd/scripts/webpack/frontend.js#L29) [practice](https://github.com/guardian/support-frontend/blob/4048812d6f0c572ba1803c0f3b3d7d78fa3823dd/support-frontend/webpack.common.js#L97) in Preact codebases is to treat React as the default and alias to `preact-compat`. Therefore I think it's best to also treat React as the default in our libraries, and let the consumers decide whether they want to alias that to Preact.

This will also mean that we can remove an [awkward workaround](https://github.com/guardian/apps-rendering/blob/94b2640fe9c8afcb3e1c05448b09e7f164ec64bc/webpack.config.ts#L59) in Apps-Rendering, where we've tried to alias in the other direction in order to accommodate atoms-rendering. Unfortunately this workaround has started to [cause us problems](https://github.com/guardian/apps-rendering/pull/1401) and we think that updating atoms-rendering to conform to the React API may solve this.

## How to test

Checks on this PR pass and atoms continue to work in DCR and AR.

The webpack upgrade on AR becomes unblocked.

Possibility of changes being required in DCR, although given that DCR already aliases `react-dom` to Preact it _should_ work fine.

### DIFF
--- a/src/lib/unifyPageContent.tsx
+++ b/src/lib/unifyPageContent.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import render from 'preact-render-to-string';
+import { renderToString } from 'react-dom/server';
 
 export const unifyPageContent = ({
     elementCss,
@@ -10,7 +10,7 @@ export const unifyPageContent = ({
     elementJs: string;
     elementHtml?: string;
 }): string =>
-    render(
+    renderToString(
         <html>
             <head>
                 <meta charSet="utf-8" />


sndrs (commented):
totally agree with this. will leave it to the team to approve, but this works well in other places, and seems a good thing to standardise on. 

(for future travellers we also all just agreed the same in the @guardian/client-side-infra meeting)

would you be up for writing a recommendation to this effect too?

nitro-marky (approved):
+1 on standardising.

OllysCoding:
A quick note on getting this to work with DCR:

With build configurations unchanged, you'll get an error:
```
Module not found: Error: Package path ./compat/server is not exported from package /dotcom-rendering/node_modules/@guardian/atoms-rendering/node_modules/preact
```

I believe this is because of how the alias is set-up in DCR's webpack config (`webpack/frontend.js`):
`'react-dom': 'preact/compat'`

Where `react-dom/server` is replaced with `preact/compat/server` which does not resolve.

The solution I found to this issue is to add a line _before_ the existing `react-dom` alias:
```
'react-dom/server': 'preact/compat'
```

There may be another format for updating this which doesn't require adding an explicit alias; however this also ensures it doesn't cause issues anywhere else in the build.

OllysCoding (approved):
Looks good to me! I don't think there's a way to do this & have it work on DCR without my referenced changes; so just worth making sure when someone comes to update it the info is there on what change they need to make.

JamieB-gu:
Good point Olly. We'd best release this as a major version.

PR #274, opened 2021-08-02
YoutubeAtom should respect adfree users

arelra:
## What does this change?

Amend `YoutubeAtom` to accept `adsConfig: { disableAds: true }` as a flag to pass to the YouTube player.

Follows method in Frontend [youtube-player](https://github.com/guardian/frontend/blob/df31b4deeb7f72303ebfa766f6a2699e2be1a293/static/src/javascripts/projects/common/modules/atoms/youtube-player.ts#L171)

## How to test

1. Log in as a subscribed user
2. Checking a DCR rendered video page such as [this one](https://www.theguardian.com/science/2021/jul/20/jeff-bezos-rocket-design-an-inquiry). It should not show an ad before the video

## Dependencies

DCR PR https://github.com/guardian/dotcom-rendering/pull/3225 is dependent on this change.

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -38,34 +38,49 @@ declare global {
 type YoutubeStateChangeEventType = { data: -1 | 0 | 1 | 2 | 3 | 5 };
 
 type EmbedConfig = {
-    adsConfig: {
-        adTagParameters: {
-            iu: string;
-            cust_params: string;
-        };
-    };
+    adsConfig:
+        | {
+              adTagParameters: {
+                  iu: string;
+                  cust_params: string;
+              };
+              disableAds?: false;
+          }
+        | {
+              disableAds: true;
+          };
 };
 
 const buildEmbedConfig = (adTargeting: AdTargetingType): EmbedConfig => {
-    return {
-        adsConfig: {
-            adTagParameters: {
-                iu: `${adTargeting.adUnit || ''}`,
-                cust_params: encodeURIComponent(
-                    constructQuery(adTargeting.customParams),
-                ),
-            },
-        },
-    };
+    switch (adTargeting.disableAds) {
+        case true:
+            return {
+                adsConfig: {
+                    disableAds: true,
+                },
+            };
+        case false:
+        case undefined:
+            return {
+                adsConfig: {
+                    adTagParameters: {
+                        iu: `${adTargeting.adUnit || ''}`,
+                        cust_params: encodeURIComponent(
+                            constructQuery(adTargeting.customParams || {}),
+                        ),
+                    },
+                },
+            };
+    }
 };
 
-const constructQuery = (query: { [key: string]: string }): string =>
+const constructQuery = (query: { [key: string]: unknown }): string =>
     Object.keys(query)
         .map((param: string) => {
             const value = query[param];
             const queryValue = Array.isArray(value)
                 ? value.map((v) => encodeURIComponent(v)).join(',')
-                : encodeURIComponent(value);
+                : encodeURIComponent(String(value));
             return `${param}=${queryValue}`;
         })
         .join('&');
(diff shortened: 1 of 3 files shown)


src/YoutubeAtom.tsx
```diff
@@ -39,20 +39,28 @@ type YoutubeStateChangeEventType = { data: -1 | 0 | 1 | 2 | 3 | 5 };
 
 type EmbedConfig = {
     adsConfig: {
-        adTagParameters: {
+        adTagParameters?: {
             iu: string;
             cust_params: string;
         };
+        disableAds?: boolean;
     };
```
mxdvl, line 46: I think this type should be either containing `adTagParameter` or `disableAds: true`.

src/types.ts
```diff
@@ -1,9 +1,10 @@
 import { Theme } from '@guardian/types';
 
 export type AdTargetingType = {
-    adUnit: string;
+    adUnit?: string;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    customParams: { [key: string]: any };
+    customParams?: { [key: string]: any };
```
mxdvl, line 6: `Record<string, unknown>` ?

mxdvl (commented):
Great catch. Maybe the type can be narrowed a little.

src/types.ts
arelra: This was the existing type but I agree a Record with an unknown type would be better

src/YoutubeAtom.tsx
mxdvl: For example, in [`prebid.ts`](https://github.com/guardian/frontend/blob/5b25cc1996b4bd87a60a853e7e74820674457e0e/static/src/javascripts/projects/commercial/modules/header-bidding/prebid/prebid.ts#L29-L49):

```ts
type ConsentManagement =
	| {
			gdpr: GDPRConfig;
	  }
	| {
			usp: USPConfig;
	  };

type UserSync =
	| {
			syncsPerBidder: number;
			filterSettings: {
				all: {
					bidders: string;
					filter: string;
				};
			};
	  }
	| {
			syncEnabled: false;
	  };
```

mxdvl: Here’s a [nice example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogtgIwgEwMIHsB2AzAlgcygF4oBvAWACgoaoBDZAZwxwIC4yrbv7kAVOvgAKdAE504EYBFGMOFajyW4Arh0bBRuTPgDcXJTwDGKjQH0wYiXKgatO-YqUBfRwEg3yXIzoIANhAAgkwA-BzYdH6MEI7czlAAPpyUHl4+-kFMHJoqMVRurlSFlFRGWBpQAG6RuGhYePgc8Eh1rIQkCm4MzPXsyR4A9ANQab4BwTYRURAANFBDUEZ0mFBIUOhwuMDSyPld-IIi4pLSsvJ7bqocAESMG1IAFtr41zMXJuaWxzbX6MAPMjM-2erz2zny4MokNK5WAUG01T8tRYDSaiBQKIIxH63UxjX6nm8Y0yNhysz2DAEwisJxkNk6HiuUFu92BOlBKQ8H2AFhpPz+ANEQKe7LenMhBSKQA)

```ts
type EmbedConfig = {
    adsConfig: {
        adTagParameters: {
            iu: string;
            cust_params: string;
        };
		disableAds?: false;
    } | {
		disableAds: true;
	};
};

const validConfig: EmbedConfig = {
	adsConfig: {
		// disableAds: false, // can be omitted
		adTagParameters: {
			iu: "something",
			cust_params: "other_thing",
		}
	}
}

const invalidConfig: EmbedConfig = {
	adsConfig: {
		disableAds: true,
		adTagParameters: {
			iu: "something",
			cust_params: "other_thing",
		}
	}
}
```

shtukas (approved):

mxdvl (approved):

PR #281, opened 2021-09-06
Temporary fix for Knowledge quizzes

nitro-marky:
## What does this change?
This is a temporary fix while the root problem is found for knowledge type quizzes to display the wrong message. 

For now these messages have been turned off until the solution is found.

## Before
Screenshot 2021-09-06 at 12 21 23

## After
Screenshot 2021-09-06 at 12 22 00


OllysCoding (changes requested):
Looks like something funky has gone on with yarn locks & package locks here? This project only uses yarn so we shouldn't have a package.json

nitro-marky:
We use npm on this project to publish for consumption in DCR.

OllysCoding (approved):
Looks good

PR #292, opened 2021-09-30
Update YoutubeAtom to init on render rather than via iframe so we can…

arelra:
## Accessibility

- [ ] [Tested with screen reader](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#screen-readers)
- [ ] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
- [ ] [Colour contrast passed](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#colour-contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#use-of-colour)


src/YoutubeAtom.test.tsx
```diff
@@ -19,9 +19,5 @@ describe('YoutubeAtom', () => {
         const { getByTitle } = render(atom);
 
         expect(getByTitle('My Youtube video!')).toBeInTheDocument();
-        expect(getByTitle('My Youtube video!')).toHaveAttribute(
-            'src',
-            expect.stringMatching(/https:\/\/www.youtube.com\/embed.*/),
-        );
```
OllysCoding, line 25: Is there a way we can keep this test, or add another few to make sure this functions as expected?

src/YoutubeAtom.tsx
```diff
@@ -197,121 +197,138 @@ export const YoutubeAtom = ({
...
-                src={iframeSrc}
-                // needed in order to allow `player.playVideo();` to be able to run
-                // https://stackoverflow.com/a/53298579/7378674
-                allow="autoplay"
-                tabIndex={overrideImage || posterImage ? -1 : 0}
-                allowFullScreen
-                data-atom-id={`youtube-video-${assetId}`}
-                data-atom-type="youtube"
-            />
+            {iframeSrc && (
```
oliverlloyd, line 316: Maybe a better pattern would be to use `if (!iframeSrc) return null;` to bail out of the render

oliverlloyd (commented):
So is the use case here that we want to delay rendering the video atom _until_ the `adTargetting` object has been generated on the client?

If so, I _think_ we could achieve this from within DCR. We currently server render the content and then *Hydrate* it but what you're proposing here is to only render it on the client. This is the same as using a `Portal` component in `App.tsx`. See [here](https://github.dev/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/App.tsx#L730).

But, all that said. Do we want to stop server side rendering videos? Assuming that's what's happening here. What would the impact be to UX from this?

oliverlloyd:
Is the requirement here to know information about permutive? And are we saying that can't be known on the server or is there a way to work that out higher up the stack?

Ideally, we'd be moving more things to the server and making the client lighter, even if that means cache splitting, but I'm not sure what this permutive data is to be honest

oliverlloyd:
Thanks for that detailed response, Ravi. That really helps.

We've already been talking about this offline (there's some exciting possibilities here!) but one issue I wanted to drop here so i don't forget about it.

If we don't render the iframe on the server then as well as this offloading more work onto the client (which we're actively discussing) but it also means we will invalidate client side hydration. React needs the dom tree it is hydrating to match what it was on the server. If we don't have that we could see undesired results

oliverlloyd:
> Actually I think because this change only modifies the state and render in an useEffect hook the server and initial client render should be the same?

Yes, you're right. Sorry, I' wasn't seeing it before. So what we're doing here is moving the place where we render the iFrame from the server to the client. Is this mental model about right?

### Before
| Server | Client |
| --------------- | --------------- |
| Render iFrame | Hydrate overlay |
| Render overlay | iFrame loads src|

### After
| Server | Client |
| --------------- | --------------- |
|Render overlay | Hydrate overlay |
| | Render iFrame |
| |iFrame loads src|

Given the client side actions will happen in parallel (I presume) could we have the situation where the overlay is hydrated but the iframe still hasn't loaded? Meaning any clicks will return errors

arelra:
> Is this mental model about right?

Yes thats right, very helpful thanks.

It might be helpful to others to expand on that (if I've got this right):

### Before
| Step | Server | Client | Interactive |
| ---------------| --------------- | --------------- | --------------- |
| 1. | Generate SSR YoutubeAtom (iFrame + Overlay) | | |
| 2. | | Render SSR YoutubeAtom (iFrame + Overlay) | | |
| 3. |  | iFrame loads player | player can receive clicks |
| 4. |  | HydrateOnce YoutubeAtom | overlay can send clicks |
| 5. |  | Render YoutubeAtom useEffect for eventEmitters | |


### Before
| Step | Server | Client | Interactive |
| ---------------| --------------- | --------------- | --------------- |
| 1. | Generate SSR YoutubeAtom (Overlay) | | |
| 2. | | Render SSR YoutubeAtom (Overlay) | | |
| 3. |  | HydrateOnce YoutubeAtom (waitFor=consentState) | overlay can send clicks |
| 4. |  | Render YoutubeAtom useEffect for iFrame | |
| 5. |  | iFrame loads player | player can receive clicks |
| 6. |  | Render YoutubeAtom useEffect for eventEmitters | |

> Given the client side actions will happen in parallel (I presume) could we have the situation where the overlay is hydrated but the iframe still hasn't loaded? Meaning any clicks will return errors

This could happen.

We could update the atom to 'queue' up any play presses so that when the player does initialise it can check if should start playing immediately.

idrise (approved):

src/YoutubeAtom.test.tsx
idrise: not blocking merge but completely agree

PR #298, opened 2021-11-09
Migrate atoms-rendering to source v4

jamie-lynch:
## Why?

We have published a Source v4 release candidate, and are keen for dotcom-rendering and apps-rendering to be first benefit from the improved developer ergonomics 🎉   In order to do this, we must update packages that they depend on, so as to avoid bundling both v3 and v4 of Source.

We have written an eslint plugin that has automated the majority of the upgrade. 

## What has changed?

Source now comprises 3 packages:

- `@guardian/source-foundations`: equivalent to `@guardian/src-foundations`
- `@guardian/source-react-components`: contains all components from packages that were prefixed `@guardian/src-*`
- `@guardian/source-react-components-development-kitchen`: New and experimental components

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,10 +37,9 @@
         "@guardian/commercial-core": "^0.26.0",
         "@guardian/consent-management-platform": "^8",
         "@guardian/libs": "^3.1.0",
-        "@guardian/src-button": "^3.6.0",
-        "@guardian/src-foundations": "^3.6.0",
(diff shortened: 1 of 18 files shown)


lucymonie (commented):
This is some very nice streamlining 👍🏻

OllysCoding (approved):
This looks good to me!

PR #303, opened 2021-11-22
YouTube Lite

oliverlloyd:
## What does this change?
This PR affects the way we load the you tube code to play videos. Instead of eager loading the iframe we now wait until the reader has either clicked or hovered over the poster image.

In addition, we also wait for `consentState` to exist.

## Why?
This reduces the amount of content and javascript we need to download on page render by at least 200Kb, deferring that work until needed. 

## Why do we wait for consent?
We need consent to decide the ad targeting value that we pass to YouTube. By deferring the load of the iframe we can ensure that we only make this call when we have all the required information.

## `Placeholder`?
There's an edge case where if a video does not have a poster image (rare) and we have not received the `consentState` back from CMP then we need to decide what to show. We could do one of two things:

1) We could render the video anyway but default consent state to disabled.
2) We can render a placeholder

This PR has implemented point 2 but this is open to discussion

### Showing YouTube loading on mouse over
https://user-images.githubusercontent.com/1336821/142861909-83f29cf6-8f86-4558-be12-224f2bc6206c.mov

### Simulating consent being set late
https://user-images.githubusercontent.com/1336821/142864495-7117f29d-c2e1-45b0-aeee-c5adc6812271.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,7 @@
         ]
     },
     "dependencies": {
+        "@guardian/src-foundations": "^3.12.0",
         "youtube-player": "^5.5.2"
     }
 }
--- a/package.json
+++ b/package.json
@@ -139,7 +139,6 @@
         ]
     },
     "dependencies": {
-        "@guardian/src-foundations": "^3.12.0",
         "youtube-player": "^5.5.2"
     }
 }
--- a/src/YoutubeAtom.stories.tsx
+++ b/src/YoutubeAtom.stories.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { css } from '@emotion/react';
 
 import { YoutubeAtom } from './YoutubeAtom';
@@ -9,7 +9,7 @@ export default {
     component: YoutubeAtom,
 };
 
-export const DefaultStory = (): JSX.Element => {
+export const NoConsent = (): JSX.Element => {
     return (
         <div
             css={css`
@@ -26,6 +26,33 @@ export const DefaultStory = (): JSX.Element => {
                 ]}
                 duration={252}
                 pillar={ArticlePillar.Culture}
+                height={450}
+                width={800}
+            />
+        </div>
+    );
+};
+
+export const NoOverlay = (): JSX.Element => {
+    return (
+        <div
+            css={css`
+                width: 800px;
+                margin: 25px;
+            `}
+        >
+            <YoutubeAtom
+                assetId="-ZCvZmYlQD8"
+                alt=""
+                role="inline"
+                eventEmitters={[
+                    (e) => console.log(`analytics event ${e} called`),
+                ]}
+                consentState={{}}
+                duration={252}
+                pillar={ArticlePillar.Culture}
+                height={450}
+                width={800}
             />
         </div>
     );
@@ -47,6 +74,7 @@ export const WithOverrideImage = (): JSX.Element => {
                     (e) => console.log(`analytics event ${e} called`),
                 ]}
                 duration={252}
+                consentState={{}}
                 pillar={ArticlePillar.News}
                 overrideImage={[
                     {
@@ -81,6 +109,7 @@ export const WithPosterImage = (): JSX.Element => {
                 ]}
                 pillar={ArticlePillar.Sport}
                 duration={252}
+                consentState={{}}
                 posterImage={[
                     {
                         srcSet: [
@@ -107,6 +136,8 @@ export const WithPosterImage = (): JSX.Element => {
                         ],
                     },
                 ]}
+                height={450}
+                width={800}
             />
         </div>
     );
@@ -140,6 +171,7 @@ export const WithOverlayAndPosterImage = (): JSX.Element => {
                         ],
                     },
                 ]}
+                consentState={{}}
                 posterImage={[
                     {
                         srcSet: [
@@ -166,7 +198,49 @@ export const WithOverlayAndPosterImage = (): JSX.Element => {
                         ],
                     },
                 ]}
+                height={450}
+                width={800}
             />
         </div>
+    );
+};
+
+export const GiveConsent = (): JSX.Element => {
+    const [consented, setConsented] = useState(false);
+    return (
+        <>
+            <button onClick={() => setConsented(true)}>Give consent</button>
+            <div
+                css={css`
+                    width: 800px;
+                    margin: 25px;
+                `}
+            >
+                <YoutubeAtom
+                    assetId="3jpXAMwRSu4"
+                    alt="Microscopic image of COVID"
+                    role="inline"
+                    eventEmitters={[
+                        (e) => console.log(`analytics event ${e} called`),
+                    ]}
+                    consentState={consented ? {} : undefined}
+                    duration={252}
+                    pillar={ArticlePillar.News}
+                    overrideImage={[
+                        {
+                            srcSet: [
+                                {
+                                    width: 500,
+                                    src:
+                                        'https://i.guim.co.uk/img/media/4b3808707ec341629932a9d443ff5a812cf4df14/0_309_1800_1081/master/1800.jpg?width=1200&height=630&quality=85&auto=format&fit=crop&overlay-align=bottom%2Cleft&overlay-width=100p&overlay-base64=L2ltZy9zdGF0aWMvb3ZlcmxheXMvdGctZGVmYXVsdC5wbmc&enable=upscale&s=aff4b8255693eb449f13070df88e9cac',
+                                },
+                            ],
+                        },
+                    ]}
+                    height={450}
+                    width={800}
+                />
+            </div>
+        </>
     );
 };
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -178,8 +178,9 @@ export const YoutubeAtom = ({
     // Show a placeholder if:
     // 1) We don't have an iframe source yet (probably because we don't have consent)
     // AND
-    // 2) There's no overlay to replace it with
-    const showPlaceholder = !iframeSrc && !hasOverlay;
+    // 2) There's no overlay to replace it with or the reader clicked to play but we're
+    //    still waiting on consent
+    const showPlaceholder = !iframeSrc && (!hasOverlay || hasUserLaunchedPlay);
     // Load the you tube iframe if:
     // 1) We have a source string defined (Eg. We have consent)
     // AND
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -12,6 +12,7 @@ import {
 import { SvgPlay } from '@guardian/source-react-components';
 
 import { MaintainAspectRatio } from './common/MaintainAspectRatio';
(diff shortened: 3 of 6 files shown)


arelra:
Nice!

> 1. We could render the video anyway but default consent state to disabled.
> 2. We can render a placeholder

I prefer 2) mainly in order to:
a) keep the main thread unblocked
b) wait for actual consent

Which just leaves a UX question of if a generic placeholder is acceptable.

Also a minor point is when I've done my own experiments loading in the player on a user click, it can cause a bit of jank when it loads in. I don't know if this is noticeable enough to be concerned about though.

oliverlloyd:
> Also a minor point is when I've done my own experiments loading in the player on a user click, it can cause a bit of jank when it loads in. I don't know if this is noticeable enough to be concerned about though.

You can test this by running `yarn storybook` in this repo but _hopefully_ there's none. The risk is though that the `height` and `width` passed down do not correspond to the content (or they don't exist and default to the wrong values). I _think_ this will be okay because we already have poster images that are correctly sized

src/YoutubeAtom.stories.tsx
```diff
@@ -166,7 +198,49 @@ export const WithOverlayAndPosterImage = (): JSX.Element => {
...
+                                {
+                                    width: 500,
+                                    src:
+                                        'https://i.guim.co.uk/img/media/4b3808707ec341629932a9d443ff5a812cf4df14/0_309_1800_1081/master/1800.jpg?width=1200&height=630&quality=85&auto=format&fit=crop&overlay-align=bottom%2Cleft&overlay-width=100p&overlay-base64=L2ltZy9zdGF0aWMvb3ZlcmxheXMvdGctZGVmYXVsdC5wbmc&enable=upscale&s=aff4b8255693eb449f13070df88e9cac',
+                                },
+                            ],
+                        },
+                    ]}
+                    height={450}
+                    width={800}
```
mxdvl, line 241: 450×800 isn’t the same ratio as 630×1200… 🤔

src/YoutubeAtom.tsx
```diff
@@ -164,9 +165,35 @@ export const YoutubeAtom = ({
...
+    const [hasUserHovered, setHasUserHovered] = useState<boolean>(false);
     const player = useRef<YoutubePlayerType>();
 
+    const hasOverlay = overrideImage || posterImage;
+
+    // Show the overlay if:
+    // 1) We have one
+    // AND
+    // 2) it hasn't been clicked upon
+    const showOverlay = hasOverlay && !hasUserLaunchedPlay;
```
mxdvl, line 177: You can use JSDOC for these, meaning hovering over `showOverlay` anywhere will give you this context.

src/YoutubeAtom.tsx
```diff
@@ -300,21 +346,27 @@ export const YoutubeAtom = ({
...
                         setHasUserLaunchedPlay(true);
-                        player.current && player.current.playVideo();
+                        iframeSrc &&
+                            player.current &&
+                            player.current.playVideo();
                     }}
                     onKeyDown={(e) => {
                         const spaceKey = 32;
                         const enterKey = 13;
                         if (e.keyCode === spaceKey || e.keyCode === enterKey) {
```
mxdvl, line 362: Outside the scope of this PR, but [`e.keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)

src/YoutubeAtom.tsx
```diff
@@ -164,9 +165,35 @@ export const YoutubeAtom = ({
...
+    // 2) it hasn't been clicked upon
+    const showOverlay = hasOverlay && !hasUserLaunchedPlay;
+    // Show a placeholder if:
+    // 1) We don't have an iframe source yet (probably because we don't have consent)
+    // AND
+    // 2) There's no overlay to replace it with or the reader clicked to play but we're
+    //    still waiting on consent
+    const showPlaceholder = !iframeSrc && (!hasOverlay || hasUserLaunchedPlay);
+    // Load the you tube iframe if:
+    // 1) We have a source string defined (Eg. We have consent)
```
mxdvl, line 185: Might be quite pedantic, but `in extensio` might be more accurate than `exempli gratia`. 🤷 

```suggestion
    // 1) We have a source string defined (i.e. we have consent)
```

mxdvl (commented):
This is a really interesting approach. I’ve got a couple of questions:

- I wonder what happens on touch devices where there’s no such concept of hover.
- Don’t we want to autoplay videos sometimes? Maybe this will be addressed with a subsequent PR.

Otherwise a couple of minor points below 👇

oliverlloyd:
> * I wonder what happens on touch devices where there’s no such concept of hover.

In this case it will be the click (or touch) event that triggers the content download. We could also consider using an observer to do the same thing when a video scrolls into view?

Most of our traffic is mobile so we should be favouring this use case.

> * Don’t we want to autoplay videos sometimes? Maybe this will be addressed with a subsequent PR.
We autoplay for video articles. These are not currently supported by DCR but if we wanted to do this here I think that's a simple prop. Obviously, if we do this then we lose all of these benefits.

src/YoutubeAtom.tsx
oliverlloyd: Yes! I love JSDoc, doing this

src/YoutubeAtom.tsx
oliverlloyd: Yeah, my IDE is saying the same thing. I'll have a quick look at changing this, it's easier to do these things while you're in the code

mxdvl:
```suggestion
                        if (e.code === "Space" || e.code === "Enter") {
```

oliverlloyd:
[Refactor YouTubeAtom](https://trello.com/c/bO5EfmIJ/427-refactor-youtubeatom)

oliverlloyd:
Based on @mxdvl 's comments I think a nice second round PR for this work would be an enhancement where we have two modes for how we decide when to load the iframe. One for mobile where we use scroll position, and the other for desktop where we use mouse position. 

To get us there, I would like to look at raising up some of the [helper utilities](https://github.com/guardian/dotcom-rendering/blob/afd05ea8cb0d870a13ac9e1a8f9a217af7e07640/dotcom-rendering/src/web/browser/hydration/whenVisible.ts) that are currently in [this PR](https://github.com/guardian/dotcom-rendering/pull/3629). A suite of functions like `whenIdle`, `whenVisible`, `whenHovered`, could potentially be useful in lots of contexts.

package.json
```diff
@@ -139,6 +139,7 @@
         ]
     },
     "dependencies": {
+        "@guardian/src-foundations": "^3.12.0",
```
SiAdcock, line 142: We should be using the `@guardian/source-foundations` peerDependency

oliverlloyd: It is already a peer dep. I think this import was auto added by mistake by VSCode - I really need to make it stop doing that.

Removed now 👍

oliverlloyd:
Are we happy that this PR is okay to go through as it stands? @arelra @mxdvl @SiAdcock ?

### Considerations
- As a future task we could look at enhancing the onHover defer on mobile devices using onVisible (consider main media - use onIdle there?)
- We _don't_ need [the CMP work](https://github.com/guardian/dotcom-rendering/pull/3661) for this change to go into DCR but CMP does need this PR
- We're happy with a placeholder in the event there is 1) no poster image and 2) consent is still pending based on the fact we virtually always have a poster - @rcrphillips

mxdvl (approved):

src/common/Placeholder.tsx
```diff
@@ -0,0 +1,65 @@
...
+const BACKGROUND_COLOUR = neutral[93];
+
+type Props = {
+    height: number;
+    rootId?: string;
+    width?: number;
+    spaceBelow?: 1 | 2 | 3 | 4 | 5 | 6 | 9;
+    spaceLeft?: 1 | 2 | 3 | 4 | 5 | 6 | 9;
+    shouldShimmer?: boolean;
+};
```
mxdvl, line 15: Actually, a small suggestion to keep things DRY.

```suggestion

type Props = {
    height: number;
    rootId?: string;
    width?: number;
    spaceBelow?: keyof typeof space;
    spaceLeft?: typeof typeof space;
    shouldShimmer?: boolean;
};
```

src/YoutubeAtom.tsx
```diff
@@ -164,9 +165,50 @@ export const YoutubeAtom = ({
...
+     *      - We don't have an overlay - so we have to load the video straight away
+     *      - The user has clicked on the overlay, so load the video iframe!
+     *      - The user has moved their mouse over the overlay so lets pre load
+     *        the content
+     */
+    const loadIframe =
+        iframeSrc && (!hasOverlay || hasUserHovered || hasUserLaunchedPlay);
+
     useEffect(() => {
+        if (!consentState) return;
```
arelra, line 211: Should this also early return, if the iframe src has been set?

Just wondering about where `consentState` is truthy and `hasUserLaunched` toggles from false to true it would set the iframeSrc and reinitialise the player again?

src/common/Placeholder.tsx
oliverlloyd: I ended up removing these props because they're actually never used and it doesn't make sense to have this component be responsible for the margins around it in any case

src/YoutubeAtom.tsx
oliverlloyd: This is a really great spot! You're completely right, it's semantically wrong to set the source again if it's already been set.

I _think_ this won't cause double loads because we gate the download using `loadIframe` but the fact I have to say 'think' is a strong indication the code is too open. Fixed 👍

oliverlloyd: Hi @arelra I ended up reverting this change because it blocked the `autoplay` value being set in an  edge case.

If the reader has their mouse cursor over the video whilst it is being hydrated then the `onMouseOver` event isn't fired which means the overlay is removed but the video doesn't start playing 😐 .

Resetting the `iframeSrc` string shouldn't impact anything, it's still only loaded once

oliverlloyd: I've tested this code in isolation (via Storybook) and when loaded into DCR and I'm seeing the expected behaviour in each case.

Next steps:

1) This PR is an improvement to the page load experience in all cases but could be seen to add a delay to when videos are played on mobile. I want to add a check for device and, when mobile, use `onIdle` and/or `onVisible` to trigger the download.

2) Once point 1 is complete we're in a position to create a new build and bring into DCR for more testing on `code`

oliverlloyd:
I've tested this code in isolation (via Storybook) and when loaded into DCR and I'm seeing the expected behaviour in each case.

Next steps:

1) This PR is an improvement to the page load experience in all cases but could be seen to add a delay to when videos are played on mobile. I want to add a check for device and, when mobile, use `onIdle` and/or `onVisible` to trigger the download.

2) Once point 1 is complete we're in a position to create a new build and bring into DCR for more testing on `code`

OllysCoding (approved):
+1 From me!

oliverlloyd:
[Lite YouTube](https://trello.com/c/gWBpvKVC/465-lite-youtube)

PR #304, opened 2021-11-24
fix: Only apply `height: 100%` to interactive main media when `Immersive`

oliverlloyd:
## What does this change?
This extends the work done in [this PR](https://github.com/guardian/atoms-rendering/pull/243 ) by @liywjl to limit that fix to only apply to immersive articles.

## Why?
The 100% height fix is still needed to prevent the iframe resize from being constrained by its container and ending up choosing a height of 150px - which was the old problem. But on liveblogs the same fix has the opposite effect.

There's an argument at this point that we could review the various containers we are rendering main media in different contexts. Perhaps there's a way to rationalise things to remove the need for this and the original work. I did go down this route, it wasn't as simple as I wanted and I felt that I was starting to introduce more risk and complexity that the situation justified. Instead, by passing in `format` and using that to limit the original fix the issue is resolved.

### Before
Screenshot 2021-11-24 at 17 04 02

### After
Screenshot 2021-11-24 at 17 00 03

## Test articles used (for future me)
Standard: https://www.theguardian.com/cities/2017/feb/13/tipping-point-cities-exercise-more-harm-than-good
Liveblog: https://www.theguardian.com/australia-news/live/2021/nov/17/australia-news-live-nt-covid-corona-katherine-victoria-nsw-pandemic-legislation-protesters-flooding-forbes
Immersive: https://www.theguardian.com/science/grrlscientist/2012/aug/07/3*

*I used Mystery Bird here because I couldn't find any immersive articles with interactive main media. Chances are this will have been removed by the time any future us' are reading this. But if you need to do the same trick just delete the mystery bird main media in Composer and then paste [this url](https://api.nextgen.guardianapps.co.uk/embed/atom/interactive/interactives%2F2018%2F06%2Fbriefing-bubbles) in the input field (where it says paste embed code).

## Merci
Thank you @liywjl for helping me with my code archeology on this one 🙏

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,8 @@
         "react-dom": "^17.0.1",
         "ts-jest": "^24.3.0",
         "ts-loader": "^8.0.18",
-        "typescript": "^4.1.3"
+        "typescript": "^4.1.3",
+        "web-vitals": "^2.1.2"
     },
     "eslintConfig": {
         "root": true,
--- a/package.json
+++ b/package.json
@@ -108,7 +108,8 @@
         "@guardian/source-react-components": "^4.0.0-rc.1",
         "react": "^17.0.1",
         "react-dom": "^17.0.1",
-        "typescript": "^4.1.3"
+        "typescript": "^4.1.3",
+        "web-vitals": "^2.1.2"
     },
     "jest": {
         "testEnvironment": "jest-environment-jsdom-sixteen",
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[Fix interactive in main media in deadblogs](https://trello.com/c/RFIvRPTG/430-fix-interactive-in-main-media-in-deadblogs)

rcrphillips (approved):
Looks good to me

alinaboghiu (approved):

PR #309, opened 2021-12-01
feat: More advanced YouTube pre-load decisions based on device

oliverlloyd:
## What does this change?
This enhances the logic we have for deciding when to pre load the iframe by taking into account if the user is on a mobile device or not and also listening for the `onTouchStart` event

## What is pre loading?
A [previous PR](https://github.com/guardian/atoms-rendering/pull/303) added the logic to defer the loading of the YouTube iframe. By doing this we reduce the amount of content that gets initially downloaded on page load, improving the reader experience


oliverlloyd:
[Lite YouTube](https://trello.com/c/gWBpvKVC/465-lite-youtube)

oliverlloyd:
[Refactor YouTubeAtom](https://trello.com/c/bO5EfmIJ/427-refactor-youtubeatom)

src/YoutubeAtom.tsx
```diff
@@ -192,20 +210,31 @@ export const YoutubeAtom = ({
...
-     * - We have a source string defined (i.e. We have consent)
-     *
-     * and
-     *
-     * - One of these 3 things are true
-     *      - We don't have an overlay - so we have to load the video straight away
-     *      - The user has clicked on the overlay, so load the video iframe!
-     *      - The user has moved their mouse over the overlay so lets pre load
-     *        the content
+     * Load the YouTube iframe if:
```
mxdvl, line 213: I feel like this comment has become a bit redundant.

src/YoutubeAtom.tsx
```diff
@@ -192,20 +210,31 @@ export const YoutubeAtom = ({
...
      */
-    const loadIframe =
-        iframeSrc && (!hasOverlay || hasUserHovered || hasUserLaunchedPlay);
+    let loadIframe: boolean;
+    const isMobile = /Mobi/.test(navigator.userAgent);
```
mxdvl, line 216: I think this is more about testing touch v pointer than it is about mobile devices. Maybe the variable name should reflect that.

I know I was the one originally mentioning touch devices, but thinking about this again, I’m wondering if the best isn’t to just listen to a [`touchstart`][] event that would act as a `hover` state.

> It’s important to note that these types of optimisations based on the availability of touch have a fundamental flaw: they make assumptions about user behavior based on device capabilities. More explicitly, the example above assumes that because a device is capable of touch input, a user will in fact use touch as the only way to interact with it.
> 
> _from 2013: [Detecting touch: it's the 'why', not the 'how'](https://hacks.mozilla.org/2013/04/detecting-touch-its-the-why-not-the-how/)_

[`touchstart`]: https://developer.mozilla.org/en-US/docs/Web/API/Element/touchstart_event

package.json
```diff
@@ -68,7 +70,7 @@
         "react-dom": "^17.0.1",
         "ts-jest": "^24.3.0",
         "ts-loader": "^8.0.18",
-        "typescript": "^4.1.3",
+        "typescript": "4.1.3",
```
mxdvl, line 73: Why pin to this version? We recommend keeping the caret: https://github.com/guardian/recommendations/blob/main/dependencies.md#javascript-npm-dependencies

src/YoutubeAtom.tsx
```diff
@@ -39,6 +42,7 @@ type Props = {
     origin?: string;
     eventEmitters: ((event: VideoEventKey) => void)[];
     pillar: ArticleTheme;
+    isMainMedia: boolean;
```
mxdvl, line 45: Can this be an optional prop?

mxdvl (commented):
This is exciting. I wonder, though, adding all these heuristics for mobile is really what we want to do. There’s also some conflation of _touch-only_, _touch-capable_ and _mobile_ devices.

Could simply waiting for `touchstart` and triggering `hasUserHovered` be enough?

package.json
oliverlloyd: It's in the PR description, having the caret causes failures

src/YoutubeAtom.tsx
oliverlloyd: It could. But isn't it better for props to be required? As in, if I _know_ I have the prop I can make decisions based on it, but if it's required I need to guard against when it's missing and make assumptions

src/YoutubeAtom.tsx
oliverlloyd: That's a really interesting, and probably very valid, point. I'm going to look into this some more (thanks for the links!) and chat it through with team 👍

src/YoutubeAtom.tsx
oliverlloyd: Removed 👍

src/YoutubeAtom.tsx
oliverlloyd: I discussed this with @OllysCoding and we both agreed your points are spot on @mxdvl and so I've refactored this PR to use `onTouchStart` in addition to `onHover`. The result of this is a my previously rather large PR is now basically a one-liner.

It's true that this means we sometimes delay the point at which a video is downloaded but as per your points and the linked articles I don't think we were accurately making the decisions about when to do this before and waiting until the user actually interacts is more accurate.

OllysCoding (approved):

mxdvl (commented):
That’s been a fun journey!

PR #325, opened 2022-02-02
Optional tracking handlers

oliverlloyd:
## What does this change?
This makes the tracking handlers used by some atoms optional

## Why?
We allow these handlers to be passed in in case different platforms want to track things differently. This change still allows this but adds a default so the handler is no longer required

### DIFF
--- a/src/lib/ophan.ts
+++ b/src/lib/ophan.ts
@@ -1,0 +1,46 @@
+import type { OphanComponentEvent } from '@guardian/libs';
+
+declare global {
+    /* ~ Here, declare things that go in the global namespace, or augment
+     *~ existing declarations in the global namespace
+     */
+    interface Window {
+        guardian: {
+            ophan: {
+                setEventEmitter: () => void; // We don't currently have a custom eventEmitter on DCR - like 'mediator' in Frontend.
+                trackComponentAttention: (
+                    name: string,
+                    el: Element,
+                    visiblityThreshold: number,
+                ) => void;
+                record: () => void;
+                viewId: string;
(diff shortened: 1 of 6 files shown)


src/lib/ophan.ts
```diff
@@ -0,0 +1,46 @@
...
+                viewId: string;
+                pageViewId: string;
+            };
+        };
+    }
+}
+
+type OphanRecordFunction = (event: { [key: string]: any }) => void;
+
+const getOphanRecordFunction = (): OphanRecordFunction => {
```
oliverlloyd, line 26: This code is duplicated from DCR. Yes, this seems bad, but it's a symptom of the fact we do  not have a centralised solution for tracking, and I think in the absence of this duplication like this is the best option as it helps us build up a series of use cases ready for when we do decide to abstract them all together into a lib

DavidLawes: Will it be obvious any future updates to submitComponentEvent or getOphanRecordFunction need to be applied to both dcr and atoms-rendering? 

Maybe we could add jsdocs to the atom components to explain what the default is? If we add jsdocs to these atom components this file might be more discoverable in dcr? What do you think?

DavidLawes (approved):
I think we should try and help improve the discoverability of this replicated code (somehow). If we don't think that's an issue, or it's impractical to do so, feel free to ignore :)

src/lib/ophan.ts
oliverlloyd: 👍  I've added a comment with more explanation and a link to the DCR version of this file

PR #330, opened 2022-02-11
Fix `YoutubeAtom` Ophan and GA Events

arelra:
## What does this change?

Currently `YoutubeAtom` does not fire it's `EventEmitters` i.e. Ophan and GA. 

The root cause of this is that we are loading the YouTube iFrame twice.

This is because the `setIframeSrc` useEffect fires twice:
-  the first time when `loadIframe` becomes true (when `interactionStarted`)
-  the second time is when `hasUserLaunchedPlay` toggles  (when the user clicks play).

The second iFrame load causes the first player instance to be removed, taking the attached player and all the listeners with it.

The temporary fix is to not call `setIframeSrc` a second time.

It is a temporary fix as we plan to refactor this component to simplify the internal state, logic and separate the player from the UI.

We don't load the iFrame twice.

Ophan and GA progress events should fire: `play`, `25`, `50`, `75`, `end`

### Before:

IFrames:

Screenshot 2022-02-11 at 00 52 27-ci-build

No Ophan events

### After:

IFrame:

Screenshot 2022-02-11 at 01 13 28-ci-fix

Ophan events:

Screenshot 2022-02-11 at 01 20 52-dev-fix

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -220,6 +220,9 @@ export const YoutubeAtom = ({
 
         // We don't want to ever load the iframe until we know the reader's consent preferences
         if (!consentState) return;
+        // We don't want to reset the iframeSrc if it has already been set
+        // Resetting the iframe will wipe the player state and all listeners
+        if (iframeSrc) return;
 
         const adsConfig: AdsConfig =
             !adTargeting || adTargeting.disableAds


DavidLawes (approved):
🎉

oliverlloyd (approved):
A very small change but this is backed up by some very impressive team investigation work!

PR #332, opened 2022-02-14
Remove interaction optimisation on `YouTubeAtom`

oliverlloyd:
## What does this change?
This PR removes an optimisation that existed on the `YouTubeAtom` to download the youtube files early when the mouseover event fired.

## Why?
The existence of this optimisation is a problem for two reasons:

1. It is complicating state management and causing bugs. In some situations we're seeing videos not being played (especially in conjunction with a [recent PR ](https://github.com/guardian/atoms-rendering/pull/330)to prevent the iframe being loaded twice).
2. We intend to refactor this component and simpler state will be easier to reason about

This means that, in those cases where the optimisation was triggered (desktop, mainly) we're now asking the user to wait longer for the content to play. In most situations the time between `mouseover` and `click` is so small this will make little difference but where it does the flow already has a spinner incorporated as part of YT's ui so any delay is mitigated.

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -162,9 +162,7 @@ export const YoutubeAtom = ({
     pillar,
 }: Props): JSX.Element => {
     const [iframeSrc, setIframeSrc] = useState<string | undefined>(undefined);
-    const [hasUserLaunchedPlay, setHasUserLaunchedPlay] = useState<boolean>(
-        false,
-    );
+    const [overlayClicked, setOverlayClicked] = useState<boolean>(false);
     const player = useRef<YoutubePlayerType>();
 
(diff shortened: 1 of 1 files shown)


mxdvl (commented):
Have we tested this in incognito? There's some heuristics about when a video is allowed to autoplay, which may mean new users have to click twice

mxdvl (approved):
Just realised my question, while valid, would be an improvement, this cannot be a regression.

Proceed!

oliverlloyd:
> Just realised my question, while valid, would be an improvement, this cannot be a regression.
> 
> Proceed!

I was about to reply. So actually the whole concept of autoplay is about to be removed anyway (in the next piece of work) so it's kind of moot

PR #333, opened 2022-02-15
refactor: use URLSearchParams to create search params

mxdvl:
## What does this change?

Use the platform! We can get rid of `encodeURIComponent`.

Easier for people to feel confident making changes.

If the string was formatted differently, we might have issues with Youtube atoms. Hopefully these would be caught by existing tests.

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -227,13 +227,20 @@ export const YoutubeAtom = ({
                       adTargeting.customParams,
                       consentState,
                   );
-        const embedConfig = encodeURIComponent(JSON.stringify({ adsConfig }));
-        const originString = origin
-            ? `&origin=${encodeURIComponent(origin)}`
-            : '';
-        setIframeSrc(
-            `https://www.youtube.com/embed/${assetId}?embed_config=${embedConfig}&enablejsapi=1&widgetid=1&modestbranding=1${originString}&autoplay=1`,
+
+        const params = new URLSearchParams({
+            ...(origin ? { origin: encodeURIComponent(origin) } : undefined),
+            autoplay: '1',
+            embed_config: encodeURIComponent(JSON.stringify({ adsConfig })),
+            enablejsapi: '1',
+            modestbranding: '1',
+            widgetid: '1',
+        });
+        const src = new URL(
+            `/embed/${assetId}?${params.toString()}`,
+            'https://www.youtube.com',
         );
+        setIframeSrc(src.toString());
     }, [consentState, overlayClicked]);
 
     useEffect(() => {
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -229,16 +229,16 @@ export const YoutubeAtom = ({
                   );
 
         const params = new URLSearchParams({
-            ...(origin ? { origin: encodeURIComponent(origin) } : undefined),
+            ...(origin ? { origin } : undefined),
             autoplay: '1',
-            embed_config: encodeURIComponent(JSON.stringify({ adsConfig })),
+            embed_config: JSON.stringify({ adsConfig }),
             enablejsapi: '1',
             modestbranding: '1',
             widgetid: '1',
         });
         const src = new URL(
             `/embed/${assetId}?${params.toString()}`,
-            'https://www.youtube.com',
+            'https://www.youtube.com/',
         );
         setIframeSrc(src.toString());
     }, [consentState, overlayClicked]);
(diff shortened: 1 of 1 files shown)


src/YoutubeAtom.tsx
```diff
@@ -227,13 +227,20 @@ export const YoutubeAtom = ({
...
                   );
-        const embedConfig = encodeURIComponent(JSON.stringify({ adsConfig }));
-        const originString = origin
-            ? `&origin=${encodeURIComponent(origin)}`
-            : '';
-        setIframeSrc(
-            `https://www.youtube.com/embed/${assetId}?embed_config=${embedConfig}&enablejsapi=1&widgetid=1&modestbranding=1${originString}&autoplay=1`,
+
+        const params = new URLSearchParams({
+            ...(origin ? { origin } : undefined),
```
jamesgorrie, line 232: I'm a little lost here if `origin` is `undefined` you are going to spread on `undefined` which is invalid? Interesting as TypeScript isn't saying so, so is it my understanding that's wrong?

See: https://www.typescriptlang.org/play?#code/MYewdgzgLgBFBcMBKBTUAnAJgHmuglmAOYA0MehRAfDAD4wCuYmKAZoSpjALyPNsdMAKFCRYRHjADeMAHTyAFLAD8cGIiYt2YTgEoYAXyEjwEEABsUs8yCJLdQA

jamesgorrie (approved):
So much nicer, but really confused about the spread on `undefined`?

src/YoutubeAtom.tsx
mxdvl: Aah… weird it’s not picking it up. Fixed by spreading to `{}`.

src/YoutubeAtom.tsx
```diff
@@ -227,13 +227,20 @@ export const YoutubeAtom = ({
...
                   );
-        const embedConfig = encodeURIComponent(JSON.stringify({ adsConfig }));
-        const originString = origin
-            ? `&origin=${encodeURIComponent(origin)}`
-            : '';
-        setIframeSrc(
-            `https://www.youtube.com/embed/${assetId}?embed_config=${embedConfig}&enablejsapi=1&widgetid=1&modestbranding=1${originString}&autoplay=1`,
+
+        const params = new URLSearchParams({
+            ...(origin ? { origin } : {}),
```
jamesgorrie, line 232:
```suggestion
            ...(origin ?? {}),
```

jamesgorrie (approved):
Still a 👍 from me

src/YoutubeAtom.tsx
jamesgorrie: Wait this doesn't work! Never mind.

mxdvl: I think this won’t lead to our expected result

```ts
const origin = "https://www.example.com/";
console.log({ ...{ origin : origin } }); // { origin: "https://www.example.com/" }
// if you have a const in scope that matches a key, you can omit it in the below syntax, which is what I’m doing
console.log({ ...{ origin } }); // { origin: "https://www.example.com/" }

console.log({ ...origin }); // { "0": "h", "1": "t", "2": "t", "1": "p", … }
```

src/YoutubeAtom.tsx
```diff
@@ -227,13 +227,20 @@ export const YoutubeAtom = ({
...
+            autoplay: '1',
+            embed_config: JSON.stringify({ adsConfig }),
+            enablejsapi: '1',
+            modestbranding: '1',
+            widgetid: '1',
+        });
+        const src = new URL(
```
sndrs, line 239: this works with 94.79% of current guardian users, are we ok with that number?

mxdvl: For some reason I’d thought this would automatically get transpiled by Babel… but actually it’s part of our [polyfill bundle] (comes with `URLSearchParams`).

[polyfill bundle]: https://github.com/guardian/dotcom-rendering/blob/9523184ba928d242f24dbc91c441a7b9d7635aae/dotcom-rendering/src/web/server/document.tsx#L194

I’ve seen it used in:
- [`dotcom-rendering` for the `FilterKeyEventsToggle`](https://github.com/guardian/dotcom-rendering/blob/dc718e5919f04509ebaeca82bdd7ff5c95ddd672/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx#L33)
- [`dotcom-rendering` for the `ShadyPie`](https://github.com/guardian/dotcom-rendering/blob/2821e7df597573ffece867163188bb98f2d752b3/dotcom-rendering/src/web/components/ShadyPie.tsx#L1-L11)

sndrs: fantastic 🪂

mxdvl: Thanks for your diligence, always!

PR #334, opened 2022-02-17
Fix `YoutubeAtom` race conditions and refactor 

arelra:
## What does this change?

This is a refactor of `YoutubeAtom` to address the following issues:

- multiple race conditions leading to:
  - events not sent to Ophan or GA (fixed in #330 )
  - the player not auto playing or sending events in the branch: https://github.com/guardian/dotcom-rendering/pull/3887 (fixed in this PR)
- using an iFrame embed _and_ the library `youtube-player` to wrap calls to the underlying YouTube player
- over time this has become an unwieldy component that mixes concerns between `Player`, `Overlay` and `Placeholder`
- internal component state is very difficult to reason about leading to brittle code and bugs

The new version of `YoutubeAtom` has the following changes:

- decoupled UI and player
- introduces a parent component `YoutubeAtom` and child components `YoutubeAtomPlayer`, `YoutubeAtomOverlay` and `YoutubeAtomPlaceholder`
- `YoutubeAtom` coordinates message passing between sibling components
- the player no longer inserts an iFrame embed
- the player only uses the library `youtube-player` to create the player and queue up api calls
- the player now has an `onReady` listener to properly determine when the player is ready
- the player `onStateChange` and `onReady` listeners use the provided YT player passed in the event object
- simplifies `Overlay` and `Placeholder` show/hide logic
- fixes `Placeholder` height and width issues
- improves `Placeholder` 'hide' behaviour  to wait for the player to be ready
- logging to allow inspection of events and component flow via `@guardian/libs logger` e.g. `guardian.logger.subscribeTo('dotcom')`

## How to test

Run a linked version on DCR with the following scenarios:

- a video with an overlay e.g. [link](https://www.theguardian.com/uk-news/2020/dec/04/edinburgh-hit-by-thundersnow-as-sonic-boom-wakes-residents)
- a video without an overlay i.e. it should show a placeholder that is removed when the player is ready
- an article with multiple atom videos e.g. [link](https://www.theguardian.com/media/2021/jun/25/guardian-journalists-win-orwell-prize-for-video-series-anywhere-but-westminster-john-harris-john-domokos)
- turn on logging to see the events

### Before

Islands branch, player does not trigger play or send Ophan events:

https://user-images.githubusercontent.com/7014230/154463150-53cf5852-28b7-49f4-a72d-02de6e7d3940.mp4

Placeholder, not sizing correctly and hiding too soon:

https://user-images.githubusercontent.com/7014230/154464217-c9a2ab69-3922-4b27-aa86-d74c8810bbd7.mp4

### After

Islands branch, player triggers play and sends Ophan events:

https://user-images.githubusercontent.com/7014230/154464441-31178949-f8e1-4b09-b2a3-9f7a89c0f748.mp4

Placeholder, sizing correctly and hiding when the player is ready:

https://user-images.githubusercontent.com/7014230/154464788-ddbe6b0b-edbf-427e-bfee-6dacbe67e561.mp4

Logging:

youtube-atom-logging-2

## Todo

Not necessarily for this PR but:

- improve the unit test coverage where possible - however I think e2e tests are better suited to player tests
- types could be improved to only use `@types/youtube` types


src/YoutubeAtomPlaceholder.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { css } from '@emotion/react';
+
+import { Placeholder } from './common/Placeholder';
+
+export const YoutubeAtomPlaceholder = (): JSX.Element => {
```
oliverlloyd, line 6: Can we make this placeholder not shimmer and be black? I think that would be more in keeping with other media loading states

Perhaps we can also add a little transition so it doesn't appear quite so suddenly?

src/YoutubeAtom.test.tsx
```diff
@@ -21,175 +17,72 @@ const overlayImage = [
...
-            iframe.src.includes(
-                '%22cust_params%22%3A%22param1%253Dparam1%2526param2%253Dparam2',
-            ),
-        ).toBe(true);
-        // encoded consent state
-        expect(
-            iframe.src.includes('%22restrictedDataProcessor%22%3Afalse'),
-        ).toBe(true);
+        fireEvent.click(getByTestId('youtube-overlay'));
+        expect(overlay).not.toBeInTheDocument();
```
oliverlloyd, line 86: Nice. Very concise and readable 💯

src/YoutubeAtomPlayer.tsx
```diff
@@ -0,0 +1,276 @@
...
+    posterImage?: ImageSource[];
+    adTargeting?: AdTargeting;
+    consentState?: ConsentState;
+    height: number;
+    width: number;
+    title?: string;
+    origin?: string;
+    eventEmitters: ((event: VideoEventKey) => void)[];
+    autoPlay: boolean;
+    loadPlayer: boolean;
```
oliverlloyd, line 25: Similar to my other content, we could raise `loadPlayer` and `consentState` up and use them as conditionals? The player component then becomes simpler.

But again, happy to take your lead here and I also see the other side to this argument where it might be desirable to make it difficult to ever load the player without consent

src/YoutubeAtomPlayer.tsx
```diff
@@ -0,0 +1,276 @@
...
+        if (!progressEvents.hasSentPlayEvent) {
+            log('dotcom', {
+                from: loggerFrom,
+                msg: 'start play',
+                event,
+            });
+            eventEmitters.forEach((eventEmitter) => eventEmitter('play'));
+            progressEvents.hasSentPlayEvent = true;
+
+            setTimeout(() => {
```
oliverlloyd, line 84: I know this is copied code but we're using `window.setTimeout` below so maybe we can do the same here?

oliverlloyd (approved):
This is superb work! It's so great to see such carefully considered attention given to this component.

I made a few comments around the api but I don't have strong opinions on these things and am happy for you to make the call on this.

Other than that I think the change from grey to black would be good for the placeholder and maybe we can also add a story with two atoms on the same page? Just to be sure and to help with any future maintenance.

src/YoutubeAtomPlaceholder.tsx
arelra: Sure good idea, that will make the placeholder a lot more subtle.

I added a transition but commented out for now as it looks a bit jarring. Mainly because it doesn't appear for very long in the overlay->placeholder->player flow. The code is there though and we can experiment with it.

package.json
```diff
@@ -48,7 +48,7 @@
         "@types/jest": "^26.0.15",
         "@types/react": "^17.0.2",
         "@types/react-dom": "^17.0.1",
-        "@types/youtube": "^0.0.39",
+        "@types/youtube": "^0.0.46",
```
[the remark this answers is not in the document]
arelra: Just housekeeping.

It does however add types for new API features around interactive ads, which commercial want to experiment with at some point.

src/YoutubeAtomPlaceholder.tsx
oliverlloyd: Nice. And plus one for having a dedicated, inline placeholder

src/YoutubeAtomOverlay.tsx
```diff
@@ -0,0 +1,162 @@
...
+    overrideImage?: ImageSource[];
+    posterImage?: ImageSource[];
+    height: number;
+    width: number;
+    alt: string;
+    role: RoleType;
+    duration?: number; // in seconds
+    pillar: ArticleTheme;
+    overlayClicked: boolean;
+    setOverlayClicked: (flag: boolean) => void;
```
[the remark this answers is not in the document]
arelra: Good idea thanks - it makes the overlay less semantically coupled.

I've lifted `overlayClicked` up and created an `onClick` callback.

The only drawback is that because `YoutubeAtom` now decides whether to render `YoutubeAtomOverlay` (and not itself) we lose the fade-out transition that was in `YoutubeAtomOverlay`.

imo It was barely visible and if anything a little distracting as the player has its own fade-in transition when it loads.

Something to keep an eye on though.

src/YoutubeAtomPlayer.tsx
arelra: Good idea thanks - it makes the player less semantically coupled.

I've lifted `loadPlayer` and `consentState` up and created an `onReady` callback.

This also allows us to make `consentState` mandatory to `YoutubeAtomPlayer`.

The only drawback is because the callback is a dependency in an `useEffect`, I had to create the callback with `useCallback` so it is stable across renders.

arelra:
@oliverlloyd @DavidLawes 

fyi the main changes since your last review:

- lift up `consentState` to parent and make mandatory
- lift up `loadPlayer` to parent
- lift up `setOverlayClicked`
- ids to distinguish between multiple players
- test for multiple players
- do not send 'end' progress event if already sent

src/YoutubeAtomOverlay.tsx
```diff
@@ -0,0 +1,150 @@
...
+    height,
+    width,
+    alt,
+    role,
+    duration,
+    pillar,
+    onClick,
+}: Props): JSX.Element => {
+    return (
+        <>
```
oliverlloyd, line 110: I think we can remove this fragment now?

src/YoutubeAtomPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+                css`
+                    width: 100%;
+                    height: 100%;
+                    position: absolute;
+                    top: 0;
+                    left: 0;
+                    display: flex;
+                    flex-grow: 1;
+                    background-color: black;
+                    /* ${transitionOpacityIn} */
```
oliverlloyd, line 36: Are we going with not having this animation? As you say, the YT element itself animates in so perhaps it's redundant

src/YoutubeAtomPlayer.tsx
```diff
@@ -0,0 +1,318 @@
...
+    disabledAds,
+} from '@guardian/commercial-core';
+import { log } from '@guardian/libs';
+
+type Props = {
+    videoId: string;
+    overrideImage?: ImageSource[];
+    posterImage?: ImageSource[];
+    adTargeting?: AdTargeting;
+    consentState: ConsentState;
```
oliverlloyd, line 18: Having this required here makes a lot more sense

src/YoutubeAtom.tsx
```diff
@@ -161,273 +45,86 @@ export const YoutubeAtom = ({
     eventEmitters,
     pillar,
 }: Props): JSX.Element => {
-    const [iframeSrc, setIframeSrc] = useState<string | undefined>(undefined);
     const [overlayClicked, setOverlayClicked] = useState<boolean>(false);
-    const player = useRef<YoutubePlayerType>();
+    const [playerReady, setPlayerReady] = useState<boolean>(false);
```
oliverlloyd, line 49: Strong 80s computer game vibes here

src/YoutubeAtom.tsx
```diff
@@ -161,273 +45,86 @@ export const YoutubeAtom = ({
...
+     * - We don't have an overlay OR the user has clicked the overlay
      *
-     * - There's no overlay to replace it with or the reader clicked to play but we're
-     * still waiting on consent
+     * AND
      *
+     * - The player is not ready
      */
-    const showPlaceholder = !iframeSrc && (!hasOverlay || overlayClicked);
+    const showPlaceholder = (!hasOverlay || overlayClicked) && !playerReady;
```
oliverlloyd, line 73: This is nice (using `playerReady` here). It makes sense having this wait controlled like this

src/YoutubeAtom.tsx
```diff
@@ -161,273 +45,86 @@ export const YoutubeAtom = ({
...
-                    data-atom-type="youtube"
+                    width={width}
+                    title={title}
+                    origin={origin}
+                    eventEmitters={eventEmitters}
+                    /**
+                     * If there is an overlay we want to autoplay
+                     * If there is not an overlay the user will use the YouTube player UI to play
+                     */
+                    autoPlay={hasOverlay}
```
oliverlloyd, line 109: 😻

oliverlloyd (approved):
This is really great! The code reads really well and is so much easier to reason about now

src/YoutubeAtomPlaceholder.tsx
arelra: Yep will remove.

src/YoutubeAtom.tsx
```diff
@@ -161,273 +45,86 @@ export const YoutubeAtom = ({
...
-                    },
-                );
-
-            return () => {
-                listener && player.current && player.current.off(listener);
-            };
-        }
-    }, [eventEmitters, loadIframe]);
+    /**
+     * Create a stable callback as it will be a useEffect dependency
```
DavidLawes, line 87: Could you help me understand this in a bit more detail please? What I think I understand so far:
- usecallback ensures we maintain the same function reference between renderings
- there are no dependencies for this usecallback so the memoized function will never be updated
- we pass playerReadyCallback as a prop to the YoutubeAtomPlayer
- onReady is invoked inside the ready listener

^^ what would be the impact of not using usecallback?

DavidLawes:
> improve the unit test coverage where possible - however I think e2e tests are better suited to player tests

I might be mis-reading, are we suggesting increasing both unit tests and e2e tests? Is this in addition to the cypress tests running inside dcr?

src/YoutubeAtom.tsx
arelra: Sure it is one of those non-obvious, rough edges with `useEffect`.

Oh and sorry the comment could be more specific - it's actually referring to `YoutubeAtomPlayer`'s `useEffect`. I'll update it.

So in `YoutubeAtom` we have the `playerReady` state.

We want to somehow have `YoutubeAtomPlayer` invoke `setPlayerReady` so `YoutubeAtom` knows when to stop showing the overlay or placeholder.

The issue is that if we passed the callback as `onReady={() => setPlayerReady(true)}` that will create a new function on every re-render of `YoutubePlayer`.

That is fine in most cases (e.g. `setOverlayClicked()`), however because we pass that function as a dependency to the `useEffect` within `YoutubeAtomPlayer` the callback function must be 'stable'. Otherwise within the `useEffect` you might get a 'stale' version of the function.

The impact being when the library `youtube-player` tries to call `onReady()` it has a reference to a function which no longer exists. I actually saw this happen - a quick test is to comment out `onReady` from the dependency array and pass in a regular callback and you should see an error in the console.

These links might explain the issue better:
https://dmitripavlutin.com/dont-overuse-react-usecallback/#2-the-purpose-of-usecallback
https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

arelra:
> > improve the unit test coverage where possible - however I think e2e tests are better suited to player tests
> 
> I might be mis-reading, are we suggesting increasing both unit tests and e2e tests? Is this in addition to the cypress tests running inside dcr?

Yes it was more of a TODO comment that we could improve the atom e2e tests in DCR to cover all scenarios i.e. placeholder only, multiple videos, checking for GA events etc.

I'm not sure how much mileage there is in extending the unit tests to cover the player. I imagine you would have to mock the player library to such an extent it undermines the value of the test, in which case e2e tests are better.

PR #337, opened 2022-03-03
Add video controls prop to YoutubeAtom

joecowton1:
## What does this change?

As part of the work to implement sticky videos in DCR, it's become apparent that we need to be able to stop Youtube videos when we close the `StickyVideo` wrapper.

This PR is a quick spike of one approach we could take in order to allow us to control the video via an additional `videoControls` prop on the `YoutubeAtom`. The idea here is to start a conversation around the best way to approach the problem.

### Which video controls to expose?
While we only currently need  to expose `stopVideo` does it make sense to account for future use cases that may need the `playVideo` and `pauseVideo` controls? The list of available controls is [here](https://developers.google.com/youtube/iframe_api_reference#Playback_controls): 

```
export type VideoControls = 'play' | 'stop' | 'pause';
```
### Is a `videoControls` state the best interface to use?
Currently the storybook example uses a `videoControls` state which when updated passes a prop down to the `YoutubeAtomPlayer`, which triggers a useEffect which in turn calls `player.current?.stopVideo()` etc. While this works, it requires the `videoControls` state to be reset to `undefined` after each state update - is there a better way of doing this?

### Listeners and events vs React state

@arelra has a detailed explanation of the issues we encountered here: https://github.com/guardian/atoms-rendering/pull/337#discussion_r820579947 

## How to test

If running locally, the storybook `Video Controls` story is the place to look.

## Images

https://user-images.githubusercontent.com/77005274/156595858-7a2b9e3b-a2ad-4f7a-9887-e4c7a7219dda.mov

## Accessibility


joecowton1:
A little brain dump before i wrap up for the day:

I think the reason this crashes with the useEffect cleanup is that using a `videoControls` prop to control the video playback is clashing with the event based (`onReady`, `onStateChange`) setup that currently exists within the atom.

When we update the `videoControls` prop at story level, that triggers a render cycle and the `YoutubeAtomPlayer` component crashes regardless of whether the prop is being passed through to it or not. 

I think the problem lies in that the iframe is being updated `onStateChange` or `onReady` whereas the component is updated on prop change. How we get around this i'm not sure 🤔

src/YoutubeAtomPlayer.tsx
```diff
@@ -273,16 +284,14 @@ export const YoutubeAtomPlayer = ({
                     }
                 };
 
-                player.current && player.current.on('ready', readyListener);
+                const readyListener = player.current?.on(
```
joecowton1, line 287: @arelra I noticed in the docs for [youtube-player](https://github.com/gajus/youtube-player) that they create a variable for each event listener which is then used as the argument for `player.off`. Using this approach here seems to resolve the issue with crashing.

```
var listener = player.on(/* ... */);

player.off(listener);
```

joecowton1: Although this works, an issue now is that when the `videoControls` prop is changed, the cleanup runs and the listeners are not reinitiated, so the tracking events are no longer fired.

arelra: > I noticed in the docs for [youtube-player](https://github.com/gajus/youtube-player) that they create a variable for each event listener

Good catch!

arelra: > Is there a way we can reset the listeners?

I think the way the listeners currently get unregistered is incorrect and we can improve it:

#### Problem(s)

I think there's a few different issues combined here!

1. Another subtle useEffect foot-gun to add to the collection, is that useEffect cleanup functions run on _every_ useEffect trigger i.e. if a dependency has changed:
	- https://maxrozen.com/demystifying-useeffect-cleanup-function

1. Storybook components are not passing props to `YoutubeAtom` as `YoutubeBlockComponent` does:
    
    - In Storybok the `eventEmitters` prop is created for each render which causes the `initPlayer` useEffect to trigger for every re-render in Storybook.
    - When used via `YoutubeBlockComponent`, `eventEmitters` are 'stable' so the cleanup function does not trigger on re-renders.
    
1. New re-renders introduced :
	
	- Previously `YoutubeAtomPlayer` would only render once so we never saw any re-render related issues.
	- Setting `videoControls` causes x2 re-renders (first to set the control, second to reset it to undefined)
	- This is not a problem per se but in Storybook re-renders are causing the useEffect cleanup to run which is pointing to a problem with the useEffect setup.

1. Incorrect placement of listener cleanup:

	- Because of 1) and 3) I think the current placement of the listener cleanup function (carried over from ye olde component) in the `initPlayer` useEffect is incorrect as the cleanup will run on every change to one of its dependencies.
	- In practise this isn't a problem as none of its dependencies _should_ change but in Storybook they do because of the unstable `eventEmitters`.
		
#### Fix

I _think_ a 'proper' fix is to separate the listener cleanup from the `initPlayer` useEffect to ensure it only runs on component unmount.

To do this I've:

- stored the listeners on a ref
- added a separate useEffect to remove the listeners with an empty dependencies array. This should only remove the listeners on unmount.

It should now work with no console errors and event emitters should be unaffected.

I've not tested this exhaustively so please test and verify :)

src/YoutubeAtom.stories.tsx
```diff
@@ -247,7 +247,13 @@ export const GiveConsent = (): JSX.Element => {
 };
 
 export const ExternalControls = (): JSX.Element => {
-    const [videoControls, setVideoControls] = useState<VideoControls>();
+    const [videoControls, setVideoControls] = useState<VideoControls>(
+        undefined,
+    );
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
arelra: Just wondering why we need to reset to `undefined`?

I might have this wrong but if the state changes the `useEffect` should fire and poke the player accordingly, and we don't need to reset?

Also a minor thing if the `videoControls` represent the state rather than an 'action' I think it would make it easier to reason about the state?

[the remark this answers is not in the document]
joecowton1: Hmm the issue here is that (at least in Storybook) if the `videoControls` state is set to `stop`, if we recommence playing the video using the youtube controls, if we press `stop` again on the external controls the state won't update and the control will have no effect. 

I'm wondering if this is a bad sign and there's a better way of exposing the controls?

src/YoutubeAtomPlayer.tsx
joecowton1: This great, thank you! I've been doing some testing and it all works nicely 👌

src/YoutubeAtom.stories.tsx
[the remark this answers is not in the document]
arelra: Ah I see thanks for explaining!

Two options perhaps with differing tradeoffs on correctness vs simplicity:

1) To ensure the player always triggers with a 'new' video control command we could modify `YoutubeAtom` to pass something like `${videoControl}-${Date.now()}` -> `play-1646662137207 ` to generate some uniqueness. This would ensure that each video control would update the player regardless of the preceding value. It is a little hacky but quite simple.

2) Maintain internal player state in `YoutubeAtomPlayer`. The YouTube player API provides the method: `player.getPlayerState()` which returns an enum indicating it's state. We already use this for event emitters in the `onStateChangeListener`. We could also use it to set some internal `playerState` and then the dependencies to the controls useEffect would be a combination of `videoControls` and `playerState` i.e. if those two values differ then we update the player state. This is probably more 'correct' but it would be more involved than 1).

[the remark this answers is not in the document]
arelra: Yep sure makes sense 👍

joecowton1:
Closing as the description and review comments are now outdated, will reopen elsewhere.

PR #339, opened 2022-03-07
Add event emitters for `resume`, `pause` and `cued`

joecowton1:
## What does this change?

As part of the sticky video work we need to have greater awareness of the player state of youtube videos.

This PR adds the following event emitters:
- `resume` if the player has already played once but been paused
- `paused` if the player has been paused
- `cued` if the player has been stopped

Given we don't need to send these events to Ophan or GA i'm unsure as to whether we need to include the `log()` sections?

## How to test

Go to storybook, open dev tools and play / pause / stop the video.

### DIFF
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -100,6 +100,21 @@ const createOnStateChangeListener = (
             setTimeout(() => {
                 checkProgress();
             }, 3000);
+        } else {
+            log('dotcom', {
+                from: loggerFrom,
+                videoId,
+                msg: 'resume play',
+                event,
+            });
+            eventEmitters.forEach((eventEmitter) => eventEmitter('resume'));
+
+            /**
+             * Set a timeout to check progress again in the future
+             */
+            setTimeout(() => {
+                checkProgress();
+            }, 3000);
         }
 
         const checkProgress = async () => {
@@ -153,6 +168,26 @@ const createOnStateChangeListener = (
                 setTimeout(() => checkProgress(), 3000);
             }
         };
+    }
+
+    if (event.data === YT.PlayerState.PAUSED) {
+        log('dotcom', {
+            from: loggerFrom,
+            videoId,
+            msg: 'paused',
+            event,
+        });
+        eventEmitters.forEach((eventEmitter) => eventEmitter('pause'));
+    }
+
+    if (event.data === YT.PlayerState.CUED) {
+        log('dotcom', {
+            from: loggerFrom,
+            videoId,
+            msg: 'cued',
+            event,
+        });
+        eventEmitters.forEach((eventEmitter) => eventEmitter('cued'));
     }
 
     if (
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -108,13 +108,6 @@ const createOnStateChangeListener = (
                 event,
             });
             eventEmitters.forEach((eventEmitter) => eventEmitter('resume'));
-
-            /**
-             * Set a timeout to check progress again in the future
-             */
-            setTimeout(() => {
-                checkProgress();
-            }, 3000);
         }
 
         const checkProgress = async () => {
--- a/src/types.ts
+++ b/src/types.ts
@@ -157,6 +157,15 @@ export type SharingUrlsType = {
     };
 };
 
-export type VideoEventKey = 'play' | '25' | '50' | '75' | 'end' | 'skip';
+export type VideoEventKey =
+    | 'play'
+    | '25'
+    | '50'
+    | '75'
+    | 'end'
+    | 'skip'
+    | 'pause'
+    | 'resume'
+    | 'cued';
 
 export type VideoControls = 'play' | 'stop' | 'pause';


src/YoutubeAtomPlayer.tsx
```diff
@@ -155,6 +170,26 @@ const createOnStateChangeListener = (
         };
     }
 
+    if (event.data === YT.PlayerState.PAUSED) {
+        log('dotcom', {
```
joecowton1, line 174: Can we remove these sections?

src/YoutubeAtomPlayer.tsx
```diff
@@ -94,6 +94,21 @@ const createOnStateChangeListener = (
...
             progressEvents.hasSentPlayEvent = true;
 
+            /**
+             * Set a timeout to check progress again in the future
+             */
+            setTimeout(() => {
+                checkProgress();
+            }, 3000);
+        } else {
+            log('dotcom', {
```
joecowton1, line 104: And this?

src/YoutubeAtomPlayer.tsx
```diff
@@ -94,6 +94,21 @@ const createOnStateChangeListener = (
             eventEmitters.forEach((eventEmitter) => eventEmitter('play'));
             progressEvents.hasSentPlayEvent = true;
 
+            /**
+             * Set a timeout to check progress again in the future
+             */
+            setTimeout(() => {
```
marjisound, line 100: Would you mind explaining why we'd need to repeat this check? Would it not be the case that every time the video state changes, `createOnStateChangeListener` is called? Like when the user plays play again (e.g. for the second time), `createOnStateChangeListener` is called and then it emits the events?

src/YoutubeAtomPlayer.tsx
marjisound: Good point, where can we access these logs. looking at the implementation of the `log` it seems it's just doing `console.log` but not sure if it's going into any S3 or something 🤔

marjisound (commented):
Loved your PR, looks amazing 😍  just had a question about the `setTimeout`

src/YoutubeAtomPlayer.tsx
joecowton1: Ahh I hadn't realised that `checkProgress()` contains a `setTimeout` which recursively calls `checkProgress()` on line `161` so I've removed this. 

And yeah you're right i think: `createOnStateChangeListener` is called every time something updates, I was just trying to make sure we still ran `checkProgress` if we were resuming rather than playing but its not necessary.

joecowton1:
>> Good point, where can we access these logs. looking at the implementation of the log it seems it's just doing console.log but not sure if it's going into any S3 or something 🤔

Yeah i don't think we need it but we should double check!

src/YoutubeAtomPlayer.tsx
arelra: Just picking up some questions raised here:

> i think: createOnStateChangeListener is called every time something updates

We only initialise the player (and it's listeners) once in the `Initialise player useEffect`.

So `createOnStateChangeListener` is actually only called once. We ensure the player is only initialised once by checking:

```js
if (!player.current)
```

Once the player is initialised `player.current` is set so we never re-initialise the player or its listeners.

`stateChangeListener` listener will then internally keep checking progress with a `setTimeout`.

src/YoutubeAtomPlayer.tsx
arelra: The `log` function is from `@guardian/libs` which will log messages to the console  but **_only_** if you manually subscribe via:

```js
guardian.logger.subscribeTo('dotcom')
```

I personally find the logs very useful when debugging so I would vote for keeping them.

If you have it switched on the output looks like this:

youtube-atom-logging-2

arelra:
> Given we don't need to send these events to Ophan or GA i'm unsure as to whether we need to include the `log()` sections?

It might help to clarify that the lines:

```js
eventEmitters.forEach((eventEmitter) => eventEmitter('cued'));
```

are looping through all the event emitters and sending events (i.e. to Ophan and GA) so if you don't want to send events for `resume`, `pause` and `cmd` then these new event emitters are redundant?

Unless you want to selectively send different events for different emitters?

The `log` function is just a debug utility so if the event does not need to be sent anywhere then they can both be removed - i.e. the event emitter and log.

joecowton1:
> > Given we don't need to send these events to Ophan or GA i'm unsure as to whether we need to include the `log()` sections?
> 
> For the emitters it might help to clarify that the lines:
> 
> ```js
> eventEmitters.forEach((eventEmitter) => eventEmitter('event'));
> ```
> 
> will loop through _all_ the event emitters and invoke them, therefore calling the Ophan and GA emitters.
> 
> So if you don't need to send events for `resume`, `pause` and `cmd` to GA or Ophan then perhaps these new event emitter calls are redundant?
> 
> Unless you are planning to add a new emitter which should only be invoked for specific events?
> 
> In which case that will need some logic to selectively invoke the required emitter(s).

For context we need `pause`, `resume` and `cued` for the sticky video `sticky` state, I envisage we add a new `eventEmitter` callback called `playerSate` or similar in `YoutubeBlockComponent` that can control an `isPlaying` state or flag that we pass into the `StickyVideo` wrapper. Apologies if you already knew that!

I think the question I had is around whether there are any costs involved in sending extra events to Ophan or GA that aren't used?

 We absolutely could add some kind of logic to check which event emitters should be called on which events, I just quite like the current setup (api?) and how open ended it is.

src/YoutubeAtomPlayer.tsx
marjisound: Thank you both for the detailed explanation. I'm just so interested to make sure I've understood this code correctly :) Do you think the steps below are now correct?
1- createOnStateChangeListener is called once within useEffect
2- createOnStateChangeListener returns a function: `(event) => {...}`
3- the function is passed to the player.on as a callback
4- the callback function listens to all YouTube IFrame Player API events (such as YT.PlayerState.ENDED)
5- so every time user plays the video or pauses or etc. the callback function is called. 
6- but for progress events, setTimeout is used because these events don't exist on youtube api out of the box

If it's not correct, could we have a small call at some point?

src/YoutubeAtomPlayer.tsx
joecowton1: Ah very useful to know, thanks. Haven't come across this before.

src/YoutubeAtomPlayer.tsx
joecowton1: I think - as far as I understand - that is exactly correct.

arelra: Yep thats exactly correct @marjisound !

In case it helps some extra details as follows:

4) The `onStateChangeListener` is specifically called for 'stateChange' events. Correspondingly there are other event types (e.g. `ready`, `playbackRateChange`) that will fire their registered listeners. More info here:

https://developers.google.com/youtube/iframe_api_reference#Events

arelra:
> I think the question I had is around whether there are any costs involved in sending extra events to Ophan or GA 

Thanks for the explanation!

Feels more like a product question i.e. :
- should we send the new events to Ophan?
- does it matter if we do and they are not used?

If there is no hard requirement to specialise the emitters based on the events then I agree, keeping the emitters generic is the best route.

src/YoutubeAtomPlayer.tsx
marjisound: Thanks both, very useful info ❤️

src/YoutubeAtomPlayer.tsx
marjisound: Ah that's really cool thing

joecowton1:
@rtyley could you possibly help us with the above questions please?

Specifically:
- should we send the new events to Ophan?
- does it matter if we do and they are not used?

rtyley:
I think that Ophan _does_ want to know whether a video is playing, or not, on a page - Ophan's javascript has code in it that attempts to keep track of how long a user is spending actually giving attention to the page - we take the user scrolling the page as a sign of attention, but we also regard a video playing on a page as a sign of attention.

See https://github.com/guardian/ophan/issues/3391 and https://github.com/guardian/ophan/pull/3470 for some related code - can have a video chat about this if it helps?

joecowton1:
> I think that Ophan _does_ want to know whether a video is playing, or not, on a page - Ophan's javascript has code in it that attempts to keep track of how long a user is spending actually giving attention to the page - we take the user scrolling the page as a sign of attention, but we also regard a video playing on a page as a sign of attention.
> 
> See [guardian/ophan#3391](https://github.com/guardian/ophan/issues/3391) and [guardian/ophan#3470](https://github.com/guardian/ophan/pull/3470) for some related code - can have a video chat about this if it helps?

Thanks @rtyley that's very useful to know. I'd be very happy to chat about this in person - for the purposes of this PR though is it safe to assume that it would in fact be beneficial to send each of these new events to Ophan?

marjisound (approved):
Great work 👍

PR #344, opened 2022-03-16
Add `setIsPlaying` callback to `YoutubeAtom`

joecowton1:
## What does this change?

This PR adds a `setIsPlaying` prop to the `YoutubeAtomPlayer` so we can provide a callback to set an `isPlaying` state any the parent component using the `YoutubeAtom`.

Previously we had been using the `eventEmitters` so send play events to the parent component via a callback, this new prop separates the concerns of tracking and play state.

In DCR we are then able to call this like so:
```javascript
	const isPlayingCallback = useCallback(
		(x: boolean) => setIsPlaying(x),
		[],
	);
	
	[...]
	return (
	    	<YoutubeAtom
	    	    setIsPlaying={isPlayingCallback}
	    	    [...]
	    	 />
	  );
```

## How to test

Run storybook and check the console for play events.

### DIFF
--- a/src/YoutubeAtom.stories.tsx
+++ b/src/YoutubeAtom.stories.tsx
@@ -28,7 +28,7 @@ export const NoConsent = (): JSX.Element => {
                 pillar={ArticlePillar.Culture}
                 height={450}
                 width={800}
-                setIsPlaying={(e) => console.log(`video is playing: ${e}`)}
+                setIsPlaying={(x) => console.log(`video is playing: ${x}`)}
             />
         </div>
     );
@@ -54,7 +54,7 @@ export const NoOverlay = (): JSX.Element => {
                 pillar={ArticlePillar.Culture}
                 height={450}
                 width={800}
-                setIsPlaying={(e) => console.log(`video is playing: ${e}`)}
+                setIsPlaying={(x) => console.log(`video is playing: ${x}`)}
             />
         </div>
     );
@@ -89,7 +89,7 @@ export const WithOverrideImage = (): JSX.Element => {
                         ],
                     },
                 ]}
-                setIsPlaying={(e) => console.log(`video is playing: ${e}`)}
+                setIsPlaying={(x) => console.log(`video is playing: ${x}`)}
             />
         </div>
     );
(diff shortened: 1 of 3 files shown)


oliverlloyd (commented):
Looking through this I'm wondering if this change is needed, could we just use the `eventEmitters` array here?

As I understand it you need to know when the video is playing or not and you're using the YT state change events to capture this. But we're already firing callbacks when the `play`, `resume`, `stop` events happen so I think you don't need a _new_ callback to be passed in, I *think* you can either repurpose an existing eventEmitter or pass in a new one?

This would mean you would need to have the logic to decide when a video is playing or not in the parent component but I think this is going to be the case anyway

joecowton1:
> Looking through this I'm wondering if this change is needed, could we just use the `eventEmitters` array here?
> 
> As I understand it you need to know when the video is playing or not and you're using the YT state change events to capture this. But we're already firing callbacks when the `play`, `resume`, `stop` events happen so I think you don't need a _new_ callback to be passed in, I _think_ you can either repurpose an existing eventEmitter or pass in a new one?
> 
> This would mean you would need to have the logic to decide when a video is playing or not in the parent component but I think this is going to be the case anyway

I think tbh I'm in agreement with you here. This came from some debugging i did with @arelra this morning, and the idea that it would be good to separate responsibility for tracking and these play events. While implementing it I came around to the idea the using the event emitters was both more intuitive and ready made for the job.

Ravi - unless you've any major objections I think we just go with the event emitters? I've tested and it doesn't cause any more re-renders than this setup.

arelra:
> Ravi - unless you've any major objections I think we just go with the event emitters? I've tested and it doesn't cause any more re-renders than this setup.

I have no major objections! Sorry @joecowton1 I didn't see that `videoState` its own emitter.

My worry was piggy-backing the existing emitters but we're not doing that.

Please go ahead :)

joecowton1:
Great stuff, thanks both. Onward with the event emitters :)

arelra:
> Great stuff, thanks both. Onward with the event emitters :)

My brain woke up and remembered one niggle! Now we are setting `isPlaying` state this will cause `YoutubeBlockComponent` to re-render and hence the emitter functions to be recreated, which means unstable functions going to useEffects.

This probably works okay as the player won't re-initialise with new emitters but it feels a bit foot-gunny.

Not for now but we should revisit and use `useCallback` or move the emitters outside of the component to make the functions stable.

PR #345, opened 2022-03-16
YoutubeAtomPlayer - Fix listener unregister

arelra:
## What does this change?

Moves the cleanup function that unregisters player listeners to an empty dependency `useEffect` to ensure the cleanup only runs on component unmount.

Currently the cleanup will [run on every trigger](https://maxrozen.com/demystifying-useeffect-cleanup-function) of the player initialisation `useEffect`.

This _shouldn't_ happen if all dependencies are stable but even so it is not the correct place for unregistration, which should only run on unmount.


joecowton1 (approved):
Great work, thanks!

PR #357, opened 2022-03-22
Refine changesets workflow

joecowton1:
## What does this change?

Makes changesets the default release method.

Removes unused `NODE_AUTH_TOKEN` value.


SiAdcock (approved):
👌

PR #365, opened 2022-03-24
YoutubeAtom - Fix Duplicate Videos

arelra:
## What does this change?

Fixes a bug when we have duplicate videos on the same page.

This issue is more apparent on liveblogs where the same video can be in the main media position and in a block.

This is the corresponding DCR fix to the Frontend fix: https://github.com/guardian/frontend/pull/24713

### Bug

Currently we use the `videoId` for the YouTubeAtom component ids e.g. :

`<div id="youtube-player-${videoId} ...>`

This is not unique when there are multiple instances of the same video and hence the YouTube player will not mount duplicate instances.

### Fix

The fix is to combine `elementId` as provided by `DotcomRenderingDataModel` e.g. :

`<div id="youtube-player-${videoId}-${elementId} ...>`

With this the YouTube player can correctly select and mount each video instance.

### elementId

Expects `elementId` to be passed from DCR as detailed here:
https://github.com/guardian/frontend/blob/f4c8237babfa16b6efe33006a55084dbf06fe462/common/app/model/dotcomrendering/PageElement-Identifiers.md

Example `DotcomRenderingDataModel` output with generated `elementId`s (search for `YoutubeBlockElement `):
https://www.theguardian.com/media/2021/jun/25/guardian-journalists-win-orwell-prize-for-video-series-anywhere-but-westminster-john-harris-john-domokos.json?dcr=true

## Images

| Before      | After |
| ----------- | ----------- |
|   Screenshot 2022-03-24 at 11 24 08 | Screenshot 2022-03-24 at 11 24 55 |

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -57,6 +57,8 @@ export const YoutubeAtom = ({
     const [isPlaying, setIsPlaying] = useState<boolean>(false);
     const [stopVideo, setStopVideo] = useState<boolean>(false);
 
+    const uniqueId = `${videoId}-${elementId}`;
+
     /**
      * Update the isPlaying state based on video events
      *
@@ -134,8 +136,8 @@ export const YoutubeAtom = ({
             <MaintainAspectRatio height={height} width={width}>
                 {loadPlayer && consentState && (
                     <YoutubeAtomPlayer
-                        elementId={elementId}
                         videoId={videoId}
+                        uniqueId={uniqueId}
                         adTargeting={adTargeting}
                         consentState={consentState}
                         height={height}
@@ -154,8 +156,7 @@ export const YoutubeAtom = ({
                 )}
                 {showOverlay && (
                     <YoutubeAtomOverlay
-                        elementId={elementId}
-                        videoId={videoId}
+                        uniqueId={uniqueId}
                         overrideImage={overrideImage}
                         posterImage={posterImage}
                         height={height}
@@ -169,10 +170,7 @@ export const YoutubeAtom = ({
                     />
                 )}
                 {showPlaceholder && (
-                    <YoutubeAtomPlaceholder
-                        elementId={elementId}
-                        videoId={videoId}
-                    />
+                    <YoutubeAtomPlaceholder uniqueId={uniqueId} />
                 )}
             </MaintainAspectRatio>
         </YoutubeAtomSticky>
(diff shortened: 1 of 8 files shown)


src/YoutubeAtomPlaceholder.tsx
```diff
@@ -2,14 +2,17 @@ import React from 'react';
 import { css } from '@emotion/react';
 
 export const YoutubeAtomPlaceholder = ({
+    id,
     videoId,
 }: {
+    id: string;
     videoId: string;
 }): JSX.Element => {
+    const uniqueId = `youtube-placeholder-${videoId}-${id}`;
```
joecowton1, line 11: Given we're creating a `uniqueId` both here and in `YoutubeAtomPlayer` does it make sense to pull it up to the `YoutubeAtom` and passing that in rather than `id`?

src/YoutubeAtom.tsx
```diff
@@ -14,6 +14,7 @@ import type { ArticleTheme } from '@guardian/libs';
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
 
 type Props = {
+    id: string;
     assetId: string;
```
oliverlloyd, line 18: Is it worth refactoring things a bit to reduce the number of variable names we have relating to ids? Is `assetId` always the same as `videoId`? And is `id` always `elementId`?

oliverlloyd (approved):
I think this looks good, just a comment on the naming of things as I think there's scope for rationalising things a bit to help future travellers.

src/YoutubeAtom.tsx
arelra: I agree this could be better!

I think there are two questions here:

- do we require both ids?
- can we rename the props to be more meaningful?

#### Required ids

We do need both ids:

- `assetId` (locally `videoId`) for the YouTube player
- `id` for uniqueness

#### Naming

I wasn't sure what to do with the naming as the conventions come from DCR.

`id` (aka `element.id`) is the prop we get from DCR. `elementId` is more descriptive but `id` follows the [existing naming convention](https://github.com/guardian/dotcom-rendering/blob/a78a98b3001cc73be46f52078bd75153d53741a2/dotcom-rendering/src/web/lib/renderElement.tsx#L458-L512).

`assetId` is the prop we get from DCR but it is locally renamed  to `videoId` to be more meaningful in the local context.

What I can do is push up the renamed props to the `YoutubeBlockComponent` level so within `YoutubeAtom` we have consistent naming?

|DCR| YoutubeAtom naming | Proposed YoutubeBlockComponent naming |
|:---|:---|:---|
| assetId | videoId | videoId |
| id | id | elementId |

PR #367, opened 2022-03-28
Add isMainMedia prop to `YoutubeAtom`

joecowton1:
## What does this change?

Adds an `isMainMedia` prop to the `YoutubeAtom` which is in turn passed to the `YoutubeAtomSticky` component, allowing us to adjust the sticky videos position for main media videos.

Currently the sticky video aligns with the right hand margin of its original container, if this a main media video we need to add `20px` padding right.

The DCR PR to implement these changes once the new version of Atoms Rendering is released [is here](https://github.com/guardian/dotcom-rendering/pull/4426).

### Details

- added `isMainMedia` prop
- tidied up `stickyContainerStyles`
- made `shouldStick` prop optional for consistency in DCR

### Before
Screenshot 2022-03-28 at 10 06 01

### After

Screenshot 2022-03-28 at 10 04 55

### DIFF
--- a/src/YoutubeAtomSticky.tsx
+++ b/src/YoutubeAtomSticky.tsx
@@ -79,23 +79,29 @@ const stickyStyles = css`
     }
 `;
 
-const stickyContainerStyles = (height: number) => css`
-    height: ${height}px;
-    position: relative;
-    display: flex;
-    justify-content: flex-end;
+const stickyContainerStyles = (isMainMedia?: boolean) => {
+    const height = 192;
 
-    ${from.tablet} {
-        height: ${height * 2}px;
-    }
-`;
+    return css`
+        height: ${height}px;
+        position: relative;
+        display: flex;
+        justify-content: flex-end;
+        padding-right: ${isMainMedia ? '20px' : 0};
+
+        ${from.tablet} {
+            height: ${height * 2}px;
+        }
+    `;
+};
 
 type Props = {
     assetId: string;
     eventEmitters: ((event: VideoEventKey) => void)[];
-    shouldStick: boolean;
+    shouldStick?: boolean;
     onStopVideo: () => void;
     isPlaying: boolean;
+    isMainMedia?: boolean;
     children: JSX.Element;
 };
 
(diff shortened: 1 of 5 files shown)


src/YoutubeAtomSticky.tsx
```diff
@@ -79,23 +79,29 @@ const stickyStyles = css`
     }
 `;
 
-const stickyContainerStyles = (height: number) => css`
-    height: ${height}px;
-    position: relative;
-    display: flex;
-    justify-content: flex-end;
+const stickyContainerStyles = (isMainMedia?: boolean) => {
+    const height = 192;
```
abeddow91, line 83: Nice and clear

src/YoutubeAtomSticky.tsx
```diff
@@ -79,23 +79,29 @@ const stickyStyles = css`
...
-    ${from.tablet} {
-        height: ${height * 2}px;
-    }
-`;
+    return css`
+        height: ${height}px;
+        position: relative;
+        display: flex;
+        justify-content: flex-end;
+        padding-right: ${isMainMedia ? '20px' : 0};
```
abeddow91, line 90: Could we use `space[5]` from Source instead of `20px`?

abeddow91 (approved):
Nice work, one small Source change if possible. Otherwise looks great!

src/YoutubeAtomSticky.tsx
joecowton1: Yeah good spot, fixed.

alinaboghiu:
Would it be possible to find a different element on the page to always align the sticky video against? E.g always align against the main body with the same amount of padding, regardless of the original video position/type. Just a thought, not sure it's possible.

joecowton1:
> Would it be possible to find a different element on the page to always align the sticky video against? E.g always align against the main body with the same amount of padding, regardless of the original video position/type. Just a thought, not sure it's possible.

It's in interesting thought for sure, and would certainly be possible. However, the way we're positioning the sticky video currently is done purely with CSS (`position: fixed` against the parent container) and works regardless of where the atom is used. 

I think if we get into adding query selectors for specific parent containers there's a risk of unnecessary specificity, we'd also (i think) end up adding more JS which is something we've tried hard to avoid throughout.

joecowton1:
[Sticky Video: right hand padding on main media](https://trello.com/c/2TyYedya/292-sticky-video-right-hand-padding-on-main-media)

PR #370, opened 2022-03-28
Fix `YoutubeAtom` tabbing issues

joecowton1:
## What does this change?

As part of the sticky videos work, and recently highlighted by the DAC reports, we've become aware that YoutubeAtom's were difficult to use with keyboard only navigation.

This PR removes the `tabIndex` values, and makes the overlay component a `button` rather than a `div`. The result is an atom which is much easier to navigate using keyboard controls.

DAC issues:

[Active elements were found that could not be navigated to via the keyboard.](https://drive.google.com/file/d/1S8ElKJj6sRNrEu5zQNp6fGejocXKgUM-/view
)
[Missing focus indicator](https://drive.google.com/file/d/1-aPrCLPJm-7lqiwrSwyfRc1BfP138YIa/view
)

## How to test

Go to storybook `YoutubeAtom/Overlay`, tab into the component and use `enter` to start the video. The controls should be available by keyboard navigation.

## Before

https://user-images.githubusercontent.com/77005274/160433223-1f50d2ef-e127-42c7-a06d-4ac6e751b669.mov

## After

https://user-images.githubusercontent.com/77005274/160433137-95988ac3-3f34-45cd-94f4-955d02a245c3.mov

### DIFF
--- a/.changeset/neat-bees-suffer.md
+++ b/.changeset/neat-bees-suffer.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': patch
+---
+
+Make YoutubeAtom controls keyboard accesible
--- a/src/Picture.tsx
+++ b/src/Picture.tsx
@@ -8,8 +8,8 @@ type Props = {
     imageSources: ImageSource[];
     role: RoleType;
     alt: string;
-    height: string;
-    width: string;
+    height: number;
+    width: number;
     isMainMedia?: boolean;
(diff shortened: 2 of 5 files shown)


abeddow91 (approved):
Great work and thanks for the write up!

joecowton1:
[Sticky Videos: tab to video controls](https://trello.com/c/SUvAymRx/285-sticky-videos-tab-to-video-controls)

PR #371, opened 2022-03-28
Add title to `YoutubeAtom` button label and iframe

joecowton1:
## What does this change?

This PR addresses issues raised in the DAC report by adding an `aria-label` to the `YoutubeAtom` overlay `play` button, and a `title` to the iframe.

DAC Issues:
[Ensure all <frame> and <iframe> elements have valid title attribute values.
](https://drive.google.com/file/d/1ZXsQqmImEIx0cirPPRsaX1LKRCeoPZp5/view)

[For each instance of this ‘Play Video’ button, the video that it is in relation should be included within
the button text.](https://drive.google.com/file/d/1hBH7i4qJVeh6Wu_laQ5FLzmL_IZbAjIe/view)

### DIFF
--- a/src/YoutubeAtom.tsx
+++ b/src/YoutubeAtom.tsx
@@ -160,6 +160,7 @@ export const YoutubeAtom = ({
                         role={role}
                         duration={duration}
                         pillar={pillar}
+                        title={title}
                         onClick={() => setOverlayClicked(true)}
                     />
                 )}
--- a/.changeset/six-tables-retire.md
+++ b/.changeset/six-tables-retire.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': patch
+---
+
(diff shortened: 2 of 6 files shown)


abeddow91 (approved):
Nice works

src/YoutubeAtom.tsx
```diff
@@ -160,6 +160,7 @@ export const YoutubeAtom = ({
                         role={role}
                         duration={duration}
                         pillar={pillar}
+                        title={title}
```
arelra, line 163: Do we also need to update `YoutubeAtom.test.tsx`?

joecowton1: Added tests for `aria-label` and `title`

arelra (approved):
👍

PR #373, opened 2022-03-29
stop other playing videos when a video plays

marjisound:
Paired with @joecowton1 and @arelra 

## What does this change?
This PR adds logic to pause the video whenever another video is played in the same page. This way we can't have multiple videos playing simultaneously. 
 

## How to test
**Storybook:**  this feature has been tested in storybook and can be checked [here](https://5f22d2148807980022fc782e-aryeqruwej.chromatic.com/?path=/story/youtubeatom--multiple-videos)

**Manual testing in DCR CODE after publishing the package:** This change needs to be deployed and published. We can then test DCR in code using the new version of atoms-rendering by going to this [page](http://localhost:3030/Article?url=https://www.theguardian.com/world/live/2022/mar/28/russia-ukraine-war-latest-news-zelenskiy-putin-live-updates) which has three videos. 
when 1 video is already playing, if you play the other video, it should stop the previous video. 

**Cypres test in DCR:** This functionality is expected to have Cypress testing in DCR

If the test in DCR code works as we expect.

## Images

https://user-images.githubusercontent.com/15894063/161256708-d7620ea1-7d15-457a-af80-625be820df49.mov

## Accessibility

### DIFF
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -44,7 +44,7 @@ type YoutubePlayerType = {
     stopVideo: () => void;
     getCurrentTime: () => number;
     getDuration: () => number;
-    getPlayerState: () => number;
+    getPlayerState: () => Promise<YT.PlayerState>;
 };
 
 type ProgressEvents = {
@@ -76,28 +76,15 @@ const createOnStateChangeListener = (
     const player = event.target;
 
     if (event.data === YT.PlayerState.PLAYING) {
-        const youtubeIframes = document.querySelectorAll(
-            '[data-atom-type="youtube"]',
-        );
-
         /**
-         * send a 'stopVideo' postMessage to all other YoutubeAtoms
+         * Emit video play event so other components
+         * get aware when a video is played
          */
-        Array.from(youtubeIframes).forEach((iframe) => {
-            if (iframe instanceof HTMLIFrameElement) {
-                // filtering out the current component playing video
-                if (iframe.id !== `youtube-video-${videoId}`) {
-                    iframe.contentWindow?.postMessage(
-                        JSON.stringify({
-                            event: 'command',
-                            func: 'stopVideo',
-                            args: [],
-                        }),
-                        'https://www.youtube.com/',
-                    );
-                }
-            }
-        });
+        document.dispatchEvent(
+            new CustomEvent('video:play', {
+                detail: { videoId },
+            }),
+        );
 
         if (!progressEvents.hasSentPlayEvent) {
             log('dotcom', {
@@ -281,6 +268,25 @@ export const YoutubeAtomPlayer = ({
                     'stateChange',
                     stateChangeListener,
                 );
+
+                /**
+                 * add a listener for when another youtube video is played in the page
+                 */
+                const youtubePlayerListener = (
+                    event: CustomEventInit<{ videoId: string }>,
+                ): void => {
+                    if (event instanceof CustomEvent) {
+                        if (event.detail.videoId !== videoId) {
+                            const playerStatePromise = player.current?.getPlayerState();
+                            playerStatePromise?.then((state) => {
+                                if (state === YT.PlayerState.PLAYING) {
+                                    player.current?.stopVideo();
+                                }
+                            });
+                        }
+                    }
+                };
+                document.addEventListener('video:play', youtubePlayerListener);
 
                 /**
                  * Register an onReady listener
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -299,7 +299,7 @@ export const YoutubeAtomPlayer = ({
                             const playerStatePromise = player.current?.getPlayerState();
                             playerStatePromise?.then((state) => {
                                 if (state === YT.PlayerState.PLAYING) {
-                                    player.current?.stopVideo();
+                                    player.current?.pauseVideo();
                                 }
                             });
                         }
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -365,7 +365,7 @@ export const YoutubeAtomPlayer = ({
                 );
 
                 /**
-                 * Record the playerListeners so they can be unregistered on component unmount
+                 * Record the listeners using refs so they can be separately unregistered _only_ on component unmount
                  */
                 playerReadyListener &&
                     playerListeners.current.push(playerReadyListener);
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -32,7 +32,7 @@ declare global {
 
 type YoutubeCallback = (e: YT.PlayerEvent & YT.OnStateChangeEvent) => void;
 
-type PlayCustomEventDetail = { videoId: string };
+type CustomPlayEventDetail = { videoId: string };
 const customPlayEventName = 'video:play';
 
 /**
@@ -357,7 +357,7 @@ export const YoutubeAtomPlayer = ({
      * Pause the current video when another video is played on the same page
      */
     const handlePauseVideo = (
-        event: CustomEventInit<PlayCustomEventDetail>,
+        event: CustomEventInit<CustomPlayEventDetail>,
     ) => {
         if (event instanceof CustomEvent) {
             if (event.detail.videoId !== videoId) {
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -75,27 +75,31 @@ const createOnStateChangeListener = (
      */
     const player = event.target;
 
-    if (event.data === YT.PlayerState.PLAYING) {
-        const iframes = document.getElementsByTagName('iframe');
-        const atomsArray = Array.from(iframes);
+    /**
+     * the ID of the iframe
+     */
+    const atomId = `youtube-video-${videoId}`;
 
-        atomsArray.forEach((iframe) => {
-            const src = iframe.getAttribute('src');
-            if (
-                iframe &&
-                iframe.id !== `youtube-video-${videoId}` &&
-                src &&
-                src.indexOf('youtube.com/embed') !== -1
-            ) {
-                console.log(iframe.id);
-                iframe.contentWindow?.postMessage(
-                    JSON.stringify({
-                        event: 'command',
-                        func: 'stopVideo',
-                        args: [],
-                    }),
-                    '*',
-                );
+    if (event.data === YT.PlayerState.PLAYING) {
+        const youtubeIframes = document.querySelectorAll(
+            '[data-atom-type="youtube"]',
+        );
+
+        /**
+         * send a 'stopVideo' postMessage to all other YoutubeAtoms
+         */
+        Array.from(youtubeIframes).forEach((iframe) => {
+            if (iframe instanceof HTMLIFrameElement) {
+                if (iframe?.id !== atomId) {
+                    iframe.contentWindow?.postMessage(
+                        JSON.stringify({
+                            event: 'command',
+                            func: 'stopVideo',
+                            args: [],
+                        }),
+                        'https://www.youtube.com/',
+                    );
+                }
             }
         });
 
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -234,7 +234,10 @@ export const YoutubeAtomPlayer = ({
         hasSent75Event: false,
         hasSentEndEvent: false,
     });
-    const listeners = useRef<Array<YoutubeCallback>>([]);
+    const playerListeners = useRef<Array<YoutubeCallback>>([]);
+    const customListeners = useRef<
+        Record<string, (event: CustomEventInit<CustomPlayEventDetail>) => void>
+    >({});
     const id = `youtube-video-${uniqueId}`;
 
     /**
@@ -295,6 +298,32 @@ export const YoutubeAtomPlayer = ({
                 );
 
                 /**
+                 * Pause the current video when another video is played on the same page
+                 */
+                const handlePauseVideo = (
+                    event: CustomEventInit<CustomPlayEventDetail>,
+                ) => {
+                    if (event instanceof CustomEvent) {
+                        if (event.detail.videoId !== videoId) {
+                            const playerStatePromise = player.current?.getPlayerState();
+                            playerStatePromise?.then((state) => {
+                                if (state === YT.PlayerState.PLAYING) {
+                                    player.current?.pauseVideo();
+                                }
+                            });
+                        }
+                    }
+                };
+
+                /**
+                 * add listener for custom play event
+                 */
+                document.addEventListener(
+                    customPlayEventName,
+                    handlePauseVideo,
+                );
+
+                /**
                  * Register an onReady listener
                  */
                 const readyListener = (event: YT.PlayerEvent) => {
@@ -332,12 +361,14 @@ export const YoutubeAtomPlayer = ({
                 );
 
                 /**
-                 * Record the listeners so they can be unregistered on component unmount
+                 * Record the playerListeners so they can be unregistered on component unmount
                  */
                 playerReadyListener &&
-                    listeners.current.push(playerReadyListener);
+                    playerListeners.current.push(playerReadyListener);
                 playerStateChangeListener &&
-                    listeners.current.push(playerStateChangeListener);
+                    playerListeners.current.push(playerStateChangeListener);
+
+                customListeners.current[customPlayEventName] = handlePauseVideo;
             }
         },
         /**
@@ -355,35 +386,6 @@ export const YoutubeAtomPlayer = ({
             width,
         ],
     );
-
-    /**
-     * Pause the current video when another video is played on the same page
-     */
-    const handlePauseVideo = (
-        event: CustomEventInit<CustomPlayEventDetail>,
-    ) => {
-        if (event instanceof CustomEvent) {
-            if (event.detail.videoId !== videoId) {
-                const playerStatePromise = player.current?.getPlayerState();
-                playerStatePromise?.then((state) => {
-                    if (state === YT.PlayerState.PLAYING) {
-                        player.current?.pauseVideo();
-                    }
-                });
-            }
-        }
-    };
-
-    /**
-     * add listener for custom play event
-     * remove listener in clean up function
-     */
-    useEffect(() => {
-        document.addEventListener(customPlayEventName, handlePauseVideo);
-        return () => {
-            document.removeEventListener(customPlayEventName, handlePauseVideo);
-        };
-    }, []);
 
     /**
      * Player stop useEffect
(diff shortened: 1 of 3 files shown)


src/YoutubeAtomPlayer.tsx
```diff
@@ -75,7 +75,34 @@ const createOnStateChangeListener = (
      */
     const player = event.target;
 
+    /**
+     * the ID of the iframe
+     */
+    const atomId = `youtube-video-${videoId}`;
```
oliverlloyd, line 81: Can we inline this? It's only used in this one place so the abstraction is only serving the purpose of giving the value a name but I think in this case it's already clear what this is. 

Perhaps if there was a need to make things clearer you could add a comment above the place where we exclude the current video explaining what we are filtering out?

src/YoutubeAtomPlayer.tsx
```diff
@@ -75,7 +75,34 @@ const createOnStateChangeListener = (
...
+
     if (event.data === YT.PlayerState.PLAYING) {
+        const youtubeIframes = document.querySelectorAll(
+            '[data-atom-type="youtube"]',
```
oliverlloyd, line 85: 👍 this makes a lot of sense

src/YoutubeAtomPlayer.tsx
```diff
@@ -75,7 +75,34 @@ const createOnStateChangeListener = (
...
+        );
+
+        /**
+         * send a 'stopVideo' postMessage to all other YoutubeAtoms
+         */
+        Array.from(youtubeIframes).forEach((iframe) => {
+            if (iframe instanceof HTMLIFrameElement) {
+                if (iframe?.id !== atomId) {
```
oliverlloyd, line 93:
```suggestion
                if (iframe && iframe.id !== atomId) {
```

Ginormous nitpick so feel free to ignore this if you are even slightly not feeling it but maybe this reads nicer?

oliverlloyd (approved):
This is really great. I can completely understand what's happening here, excellent work!

alinaboghiu (approved):

src/YoutubeAtomPlayer.tsx
joecowton1: It might be that this isn't needed in either form - see the check on the line above

marjisound: @joecowton1 do you mean we can just get away with `iframe.id` instead of `iframe && iframe.id` or `iframe?.id`?

joecowton1: yeah I _think_ so

marjisound: yea makes sense to me 👍

abeddow91 (approved):

arelra:
Super minor - could we update the PR title to describe the new behaviour a bit more closely?

Roughly something like: `When a video plays - stop other videos playing`

But whatever wording makes sense to you!

src/YoutubeAtomPlayer.tsx
```diff
@@ -75,7 +75,34 @@ const createOnStateChangeListener = (
...
+                    iframe.contentWindow?.postMessage(
```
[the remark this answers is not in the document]
joecowton1: Thanks for the comments @arelra .

My understanding is that using `postMessage` is a fairly standard way to communicate with an iframe: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

I'm not sure using custom DOM events simplifies things, or make us less reliant on Youtube exposing a `stopVideo` function on the iframe?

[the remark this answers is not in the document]
oliverlloyd: Thanks @arelra - this is a really informative and well thought through post!

My comment won't be to the same level but I think it's a really interesting and valid point. If the contract between that post message to the iframe and the you tube player responding to it is weak then we should consider other alternatives.

@marjisound What led you to using the post message approach? I really like it (it's so simple and clean!) but it'd be great to understand it a bit more. It's possible this is inherently a tricky thing to do and that there will always be complexity or fragility somewhere but I guess this is an area that could grow so is worth a bit of digging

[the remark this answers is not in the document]
arelra: Hey @joecowton1 

Yeah I was more referring to the use of `postMessage` to stop the player - which afaics is an internal undocumented feature:

https://developers.google.com/s/results/youtube/iframe_api_reference?q=postMessage

I think DOM Events would keep the interaction between components somewhat separated in a publisher/subscriber model.

On receiving the event we could call the `player.stopVideo()` which is documented on the YouTube iFrame API:

https://developers.google.com/youtube/iframe_api_reference#Functions

src/YoutubeAtomPlayer.tsx
```diff
@@ -76,6 +76,29 @@ const createOnStateChangeListener = (
...
+         * send a 'stopVideo' postMessage to all other YoutubeAtoms
+         */
+        Array.from(youtubeIframes).forEach((iframe) => {
+            if (iframe instanceof HTMLIFrameElement) {
+                // filtering out the current component playing video
+                if (iframe.id !== `youtube-video-${videoId}`) {
+                    iframe.contentWindow?.postMessage(
+                        JSON.stringify({
+                            event: 'command',
+                            func: 'stopVideo',
```
arelra, line 93: This is more of a functional question but do we want to _stop_ the video? This would re-cue the video to the start?

Would pausing be a bit more useful to the user so that they can resume the other video where they left off if they want to?

src/YoutubeAtomPlayer.tsx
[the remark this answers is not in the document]
joecowton1: Ah I think i get it now - sorry that was a bit of pre-lunch stubbornness. 

So within the `YoutubeAtomPlayer` we'd add a `dispatchEvent` on play, and also event listener for the play event, if the play event doesn't belong to the current component then we call `stopVideo`?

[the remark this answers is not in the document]
arelra: No worries @joecowton1 it was a long comment so the clarification is good!

> So within the `YoutubeAtomPlayer` we'd add a `dispatchEvent` on play, and also event listener for the play event, if the play event doesn't belong to the current component then we call `stopVideo`?

Yep! We'll have to prove it works in real life but I think that sequence should work in principle.

src/YoutubeAtomPlayer.tsx
oliverlloyd: +1 on this. Pausing would, I think, be a better experience

src/YoutubeAtomPlayer.tsx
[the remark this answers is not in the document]
marjisound: Thanks all for your comments. We're going to explore the `dispatchEvent` solution with @arelra, hopefully it'll be a seamless work. 

@oliverlloyd The only reason I used the post message approach was that I couldn't find any other way of doing it. Reading through this stack overflow answer, I tried the post message and saw that it worked https://stackoverflow.com/a/27410439/5273833

I'll update you here how our experiment with the `dispatchEvent` goes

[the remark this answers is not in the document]
oliverlloyd: I think it's worth saying that there's nothing wrong with the `postMessage` solution. Its simplicity is a real plus!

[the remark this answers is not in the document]
marjisound: Yea it looks nice but chatting with Ravi, if I understood correctly (please correct me if I'm wrong), because youtube api doesn't have this feature documented, there's no guarantee that their api would always handle this post message for stopping the video 
```
{
    event: 'command',
    func: 'stopVideo',
    args: [],
}
```

src/YoutubeAtomPlayer.tsx
```diff
@@ -278,6 +288,25 @@ export const YoutubeAtomPlayer = ({
...
+                            const playerStatePromise = player.current?.getPlayerState();
+                            playerStatePromise?.then((state) => {
+                                if (state === YT.PlayerState.PLAYING) {
+                                    player.current?.stopVideo();
+                                }
+                            });
+                        }
+                    }
+                };
+                document.addEventListener('video:play', youtubePlayerListener);
```
joecowton1, line 308: Should this be in a useEffect so we can remove the eventListener when the component unmounts?

src/YoutubeAtomPlayer.tsx
marjisound: Thanks all for pointing this out, and thanks @joecowton1 for quickly making the PR for adding the pause functionality to the interface. 

I made the change and the play event handler is now pausing the video instead of stopping

src/YoutubeAtomPlayer.tsx
```diff
@@ -32,6 +32,9 @@ declare global {
 
 type YoutubeCallback = (e: YT.PlayerEvent & YT.OnStateChangeEvent) => void;
 
+type PlayCustomEventDetail = { videoId: string };
```
joecowton1, line 35: Sorry, nits again, but could this be `CustomPlayEventDetail`?

src/YoutubeAtomPlayer.tsx
```diff
@@ -32,6 +32,9 @@ declare global {
...
+const customPlayEventName = 'video:play';
```
joecowton1, line 36: And this just `customPlayEvent`?

src/YoutubeAtomPlayer.tsx
```diff
@@ -340,6 +353,35 @@ export const YoutubeAtomPlayer = ({
...
+    /**
+     * add listener for custom play event
+     * remove listener in clean up function
+     */
+    useEffect(() => {
+        document.addEventListener(customPlayEventName, handlePauseVideo);
+        return () => {
+            document.removeEventListener(customPlayEventName, handlePauseVideo);
+        };
+    }, []);
```
arelra, line 383: Normally you'd need to add any variable used inside the `useEffect` to its dependency array i.e. `handlePauseVideo` should be in the dependency array.

However I think we can avoid this by moving the function creation of `handlePauseVideo` inside this `useEffect`.

Because this `useEffect` will only run once then we will only create it once and the listeners too.

src/YoutubeAtomPlayer.tsx
```diff
@@ -278,6 +288,25 @@ export const YoutubeAtomPlayer = ({
                     stateChangeListener,
                 );
 
+                /**
+                 * add a listener for when another youtube video is played in the page
+                 */
+                const youtubePlayerListener = (
+                    event: CustomEventInit<{ videoId: string }>,
+                ): void => {
```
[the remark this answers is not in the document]
marjisound: Thanks for sharing the article. Removed the return type 👍

src/YoutubeAtomPlayer.tsx
```diff
@@ -340,6 +353,35 @@ export const YoutubeAtomPlayer = ({
         ],
     );
 
+    /**
+     * Pause the current video when another video is played on the same page
+     */
+    const handlePauseVideo = (
```
arelra, line 359: This function will get re-created on every render of `YoutubeAtomPlayer`.

This is okay if:
- it is not expensive to create
- it is not used by any useEffect, if it is should be made stable
- we can guarantee that there aren't re-renders which avoids any referential equality issues

I think given we are using it in a `useEffect` it will be safer to move the function creation inside the `useEffect` where it will be used.

src/YoutubeAtomPlayer.tsx
marjisound: The thing is, this const is used for just the name of the event and not the actual event. That's why I thought maybe it's more clear if we call it `customPlayEventName` rather than `customPlayEvent`

But I'm happy to change if you think `customPlayEvent` suits better here?

src/YoutubeAtomPlayer.tsx
marjisound: done 👍

src/YoutubeAtomPlayer.tsx
marjisound: Thanks @arelra for explaining. Could you please elaborate a bit on your third point about referential equality check issues in useEffect?

src/YoutubeAtomPlayer.tsx
marjisound: Thank @arelra for your patience and explanation :) If we move the function creation into the useeffect, do we not also need to add function dependancies to the dependancy array?

src/YoutubeAtomPlayer.tsx
arelra: Sure @marjisound, so when we recreate functions inside functional components like this they will get recreated on every render of that component.

That can be okay for the reasons above, however if we use that function in a `useEffect` and it is in its dependency array like:

```js
const myFunction = () => {};
useEffect(() => {
    myFunction();
}, [myFunction]);
```

Every time the component rerenders, `myFunction` is recreated and this `useEffect` will trigger because `useEffect` will see that the reference equality of `myFunction` has changed.

Normally this isn't what you want so if you pass functions to `useEffect` you need to make them stable by either:
- using `useCallback` to memoise the function
- move the function outside of the component
- if the useEffect is called once (i.e. just on component mount) move the function creation inside the useEffect itself

So in this case I think we can move the function creation inside the `useEffect` as we will only trigger it once, because the empty dependency array means it will only run on component mount.

src/YoutubeAtomPlayer.tsx
joecowton1: Yeah fair point, makes sense.

src/YoutubeAtomPlayer.tsx
joecowton1: Would the values in that function (`videoId`, `player` etc) not then need to be added to the dependancy array?

src/YoutubeAtomPlayer.tsx
arelra: Great catch! Yes you're right.

I think we would need the dependency `videoId`, but not `player.current` however as you are [not supposed to pass refs](https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780) to a as a usEffect dependency

In which case we can use `useCallback` to wrap the function creation but _then_ we (in theory) lose the specific unmount `useEffect` by adding that as a dependency.

I think we should probably switch back to the two-step listener setup and unregister that the current player listeners use? That way we can guarantee conforming to all the crazy useEffect gotchas 🙃

joecowton1: IIUC and the only value in the dependancy array is `videoId`, that's not going to change, so the cleanup won't be fired until the component un-mounts?  

The reason we had issues previously was that somewhere in the deps array of the larger useEffect within `YoutubeAtomPlayer` dependancies were updating, which was causing the cleanup to run sooner.

I think it's a lot easier for anyone coming to this project afresh if this all lives in one useEffect, and we're not pushing listeners to refs etc unless we really have to?

marjisound: @arelra and @joecowton1 thank you both for your great ideas. 

Having paired with Ravi, we moved the listener function back to the original useEffect. 

As Joe mentioned, most probably the dependancy on `videoId` wouldn't have caused issues in useEffect as it's a static value. But I think 1- to be on the safe side (avoiding all the crazy useEffect gotchas ) and 2- to be consistent with how the rest of the listeners have been registered and de-registered, it may be good to follow the same pattern of registering in one useEffect and de-registering in a separate useEffect?

Please let me know if you feel concerened about it? Or you think there could still be more improvements 🙂

src/YoutubeAtomPlayer.tsx
marjisound: Thanks a lot @arelra  for the detailed explanation. It now makes more sense to me  🙂

src/YoutubeAtomPlayer.tsx
arelra: > the only value in the dependancy array is videoId, that's not going to change, so the cleanup won't be fired until the component un-mounts

Yep you're right! However this only works as a consequence of how that particular value is changed or not changed over the lifecycle of the component.

This may not work for other variables if they are introduced later to the component (and useEffect) so I'm a bit wary of introducing a pattern that might not work correctly in call cases for all dependencies.

The current pattern of refs for the player and player listeners is already established and I personally think thats actually easier and more deterministic than trying to grok when a useEffect might fire depending on its dependencies? 

But I'm very aware that might just be me as I have been staring at this for a while :)

joecowton1: With the caveat that we've all being looking at this for a long time now, and we're probably hitting the limits of the cost-benefit of a prolonged back and forth, my current thoughts are..

> However this only works as a consequence of how that particular value is changed or not changed over the lifecycle of the component.

Is this not true of every useEffect ever? :)

> The current pattern of refs for the player and player listeners is already established and I personally think thats 
> actually easier and more deterministic than trying to grok when a useEffect might fire depending on its 
> dependencies?

I think that an individual useEffect provides a separation of concerns and is easy to understand - to me it feels as though this is separate from the player listeners for which pushing to refs is a workaround, this doesn't require that.

To me something like below is a lot less intuitive than a useEffect that adds and removes and event listener, something that's very common place in React projects everywhere. I think without context I would wonder why we needed to do this:

```javascript
      customListeners.current[customPlayEventName] = handlePauseVideo;
                
     Object.entries(customListeners.current).forEach(
                ([eventName, eventHandler]) => {
                    document.removeEventListener(eventName, eventHandler);
                },
            );
```

Anyway happy to let it to slide, that's just my take on it. It's good to throughly interrogate this things 🙂

marjisound: Valid points you're making Joe. It's also about personal preference which could look more or less intuitive. 
But one point I understood from Ravi's comment is that if we have a useEffect that is having a non-empty dependancy array even if it's a static dependancy`[videoId]`, this way, we're introducing this pattern that it's OK for this useEffect to have dependancy and make it more prone to bug in the future

arelra: > Is this not true of every useEffect ever? :)

I get your point and its totally valid but I think in the particular case of unregistering listeners we want some strong guarantees that we will only do that on onmount. If we can avoid setting potential traps for ourselves that require deep understanding of re-renders, function recreation and useEffect gotchas I feel like that's a better place to be.

> this is separate from the player listeners

Yep I can see that. It gets subjective here! To me they are all listeners and if we register and deregister them in the same way I feel that groups them together and aids understanding.

> I think without context I would wonder why we needed to do this

Thats valid perhaps we could add more detail to the comments that are already there.

I think we can all agree though that useEffect is a pain! It has a problematic API and it doesn't help devs to [fall into the pit success](https://blog.codinghorror.com/falling-into-the-pit-of-success/).

arelra (approved):
Awesome work and discussion. Thanks @marjisound and @joecowton1 !

src/YoutubeAtomPlayer.tsx
marjisound: @joecowton1 I updated the comment about adding the listeners to refs. Are you happy with the current state of the PR now?

joecowton1: Thanks @marjisound - yeah nice work 🙂

marjisound:
Thank you all for your support on this PR :)

joecowton1:
[Sticky Video: handle multiple playing videos on one page](https://trello.com/c/0FKXpoj3/252-sticky-video-handle-multiple-playing-videos-on-one-page)

PR #374, opened 2022-03-31
Fix: sticky video content layout jump 

abeddow91:
## What does this change?
Reduces the height of the sticky video container to match the video height for phablets and above to prevent a content shift when the video sticks, is closed or is reinserted into its container.

## Before
https://user-images.githubusercontent.com/20416599/161040586-e31d6a52-bdc4-4552-9dbc-e620e1c44dc5.mov

## After
https://user-images.githubusercontent.com/20416599/161040495-4cd9d52a-cf3b-4061-9380-551b894d97ed.mov

### DIFF
--- a/src/YoutubeAtomSticky.tsx
+++ b/src/YoutubeAtomSticky.tsx
@@ -80,17 +80,18 @@ const stickyStyles = css`
 `;
 
 const stickyContainerStyles = (isMainMedia?: boolean) => {
-    const height = 192;
+    const videoHeight = 192;
+    const largeVideoHeight = 349;
 
     return css`
-        height: ${height}px;
+        height: ${videoHeight}px;
         position: relative;
         display: flex;
         justify-content: flex-end;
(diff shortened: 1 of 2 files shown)


src/YoutubeAtomSticky.tsx
```diff
@@ -80,17 +80,18 @@ const stickyStyles = css`
 `;
 
 const stickyContainerStyles = (isMainMedia?: boolean) => {
-    const height = 192;
+    const videoHeight = 192;
```
joecowton1, line 83: Nit but should we just inline these values now they're only used once?

abeddow91: Yeah sounds good!

joecowton1 (approved):

joecowton1:
[Sticky Video: tiny jump on close](https://trello.com/c/VWbblYas/293-sticky-video-tiny-jump-on-close)

PR #376, opened 2022-03-31
Unstick `YoutubeAtom` on pause

joecowton1:
## What does this change?

Following [comments on this PR](https://github.com/guardian/atoms-rendering/pull/373#discussion_r839527194) we need to add a listener on the `YoutubeAtomPlayer` to emit a `pause` event when paused.

This can then be used to untsick the video when it's paused using the `playerState` callback in `YoutubeAtom`.

We've also added a `resume` event so the `isPlaying` flag is set to `true` once the video is resumed, at which point it will stick again if required.

### DIFF
--- a/.changeset/big-cougars-exercise.md
+++ b/.changeset/big-cougars-exercise.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': patch
+---
+
+Untsick YoutubeAtom on pause event


marjisound (approved):
LGTM 👍

joecowton1:
[Sticky Video: add 'pause' event handling](https://trello.com/c/toOeena8/301-sticky-video-add-pause-event-handling)

PR #383, opened 2022-04-04
fix dispatch play event when duplicate videos exist in page

marjisound:
## What does this change?
This PR fixes a bug on play dispatch event for when there are duplicate youtube videos in one page. 

As a result of this bug, when a video exists multiple times in a page, if the first one is already playing, and user plays the second one, the first one was not being paused, while the expected behaviour is that playing any video in the page is supposed to pause any other playing video. 

## How to test

https://user-images.githubusercontent.com/15894063/161565631-a053d22c-f89b-4a1f-aa53-c3a6c4cb4ac9.mov

This change will be tested in DCR code before going to prod

## Accessibility

### DIFF
--- a/.changeset/cool-planets-smash.md
+++ b/.changeset/cool-planets-smash.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': patch
+---
+
+fix dispatch play event on duplicate videos


marjisound:
> 👍 Do we need a changeset?

Just did a patch changeset :)

PR #394, opened 2022-04-07
YoutubeAtomPlayer `useOnMount` custom hook

arelra:
## What does this change?

Introduces a new custom hook `useOnMount`.

Positives:

- declarative `onMount`
- abstracts away thinking about re-renders, useEffect triggers, mount and unmounting
- enforces the effect to only run on mount i.e. we could use useEffect with `[]` but it's open to misuse and re-triggering
- allows a cleanup function to be returned in a more idiomatic 'unmount' way
- allows the listener refs to be removed and the separate unmount useEffect

Negatives:

- abstracts away thinking about re-renders, triggers, mount and unmounting :) . I think we've learnt a great deal by having to deal with these concepts. Perhaps we are shielding future developers from that learning curve? Is this a negative?
- dependencies are less explicit by the fact that they are captured by the mount function and no longer declared

@marjisound and @joecowton1 I am not sure whether this is a useful abstraction so I'd be interested in your feedback !

### DIFF
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -354,9 +354,13 @@ export const YoutubeAtomPlayer = ({
         const playerReadyListener = player.current?.on('ready', readyListener);
 
         /**
-         * Return the cleanup effect that will run _only_ on unmount via useOnMount API
+         * Return the cleanup function that will run _only_ on unmount via useOnMount API
          */
         return () => {
+            log('dotcom', {
+                from: 'YoutubeAtomPlayer cleanup',
+                videoId,
+            });
             playerReadyListener && player.current?.off(playerReadyListener);
             playerStateChangeListener &&
                 player.current?.off(playerStateChangeListener);
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef } from 'react';
 import YouTubePlayer from 'youtube-player';
+import { useOnMount } from './lib/useOnMount';
 
 import type { AdTargeting, VideoEventKey } from './types';
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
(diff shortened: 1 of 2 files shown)


joecowton1:
Thanks for giving this a go @arelra ! 

I'm in two minds about it - on the one hand, as you say, it allows us to be more idiomatic (which i'm very much in favour of) and removes some potential pain and complexity for future developers, on the other hand I agree that each of our understandings have benefitted greatly from some of the struggle involved in getting this component to behave as we wanted.

Is wrapping a whole component in a `useEffect` going to make sense to anyone coming to this further down the road, and how important is that future developers actually do understand some of the pitfalls that have lead us to this point? 

If it turns out to be more confusing then I'd say maybe we leave things as they are, but in all honesty I really don't know. I'm interested to hear what others think :)

src/YoutubeAtomPlayer.tsx
```diff
@@ -237,164 +238,151 @@ export const YoutubeAtomPlayer = ({
...
-    >({});
     const id = `youtube-video-${uniqueId}`;
 
     /**
-     * Initialise player useEffect
+     * Initialise player
+     *
+     * Please refer to the useOnMount API
+     *
+     * The following dependencies are used with the values set at component mount:
```
marjisound, line 248: This looks really good. Would you mind explaining how the dependancies are used while we're not adding them to any dependancy array?

marjisound, line 248: This looks really good. Would you mind explaining how the dependancies are used while we're not adding them to any dependancy array?

arelra:
> Is wrapping a whole component in a `useEffect` going to make sense to anyone coming to this further down the road, and how important is that future developers actually do understand some of the pitfalls that have lead us to this point?

Oh it's actually just wrapping the useEffect rather than the component itself. That is a fairly standard method of creating custom hooks so hopefully not too out of the ordinary.

I agree we are removing learning opportunities but on the plus side by removing sharp edges hopefully the learning curve will be a little gentler and the code easier to work with 🤞 
 
> If it turns out to be more confusing then I'd say maybe we leave things as they are, but in all honesty I really don't know. I'm interested to hear what others think :)

I'm leaning towards this being a net positive:

- it is declarative i.e. it says (and enforces) that this function runs on mount and it returns an unmount cleanup function, rather than having to parse the useEffect to determine - and potentially make mistakes with the dependencies.
- DCR also has a bunch of custom hooks to abstract useEffect (e.g. [useOnce](https://github.com/guardian/dotcom-rendering/blob/66549659a6bf45c5ddddf09584e6b6f283ac00d8/dotcom-rendering/src/web/lib/useOnce.ts)) so I think it isn't anything out of the ordinary in terms of devex

src/YoutubeAtomPlayer.tsx
arelra: Thats exactly right @marjisound :)

marjisound: That sounds great. I'd vote for using it 👍 , as it would be a guarantee when we want some code to only run at mount stage

marjisound (approved):

joecowton1:
> Oh it's actually just wrapping the useEffect rather than the component itself. That is a fairly standard method of creating custom hooks so hopefully not too out of the ordinary.

Ah yes ofc, sorry not sure why I thought otherwise!

You make a good case for it, I think with everything you've said in mind we should go for it. The test will come if anyone else comes to work on the component soon - will be interesting to see their experiences 👍

joecowton1 (approved):

arelra:
Thanks all for your comments!

fyi I'm going to park this for the moment - there are use cases I would like to test more thoroughly but its low priority against other work happening atm.

PR #400, opened 2022-04-11
Add touch overlay to mobile `YoutubeAtomSticky` component

joecowton1:
## What does this change?

We currently have a bug on mobile web where tapping on the video does not reveal the `close` button.

This PR adds a check to see if the page is being loaded on a mobile device, and if so adds a transparent overlay to allow the user to trigger the close button via a tap. This behaviour mimics that which currently exists on the live app. 

###  Web

Screenshot 2022-04-12 at 07 57 36

### Mobile

Screenshot 2022-04-12 at 07 56 43

###  Before

https://user-images.githubusercontent.com/77005274/162748190-edab41ed-27e2-4611-abae-ddea30bf2987.mov

### After

https://user-images.githubusercontent.com/77005274/162748346-0cca87bb-020c-4aa8-bc54-f914b19b5e98.mov


src/YoutubeAtomSticky.tsx
```diff
@@ -113,8 +126,11 @@ export const YoutubeAtomSticky = ({
     isMainMedia,
     children,
 }: Props): JSX.Element => {
+    const isMobile = detectMobile();
```
marjisound, line 129: Do you think it's possible and beneficial to  make `isMobile` a ref since it should be a static value and doesn't need to change? If I'm not wrong, that way we'd call `detectMobile` only once when the component is created. 
`const isMobile = useRef(detectMobile());`

src/lib/detectMobile.ts
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable no-useless-escape */
+
+// https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser
+export const detectMobile = function () {
```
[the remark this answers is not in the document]
joecowton1: Great point Marjan, I think it probably would go out of date. We've found a tiny (8kb) NPM module here that will do the same job _and_ be patched regularly.

https://www.npmjs.com/package/is-mobile

src/YoutubeAtomSticky.tsx
joecowton1: That's an interesting one! Had a long chat with @mxdvl about the advantages and disadvantages of using `refs` before realising the blindingly obvious thing to do here is to just move the function call outside of the component 🙂 

For added context though the React docs do suggest caution when using refs: https://reactjs.org/docs/refs-and-the-dom.html

When we were working on and around the `YoutubeAtomPlayer` there were a number of reasons to use refs, but it's good to remember that React is meant to be reactive - if we use refs everywhere while it might save re-renders it also side-steps some deliberate design features in the way React is written.

package.json
```diff
@@ -64,6 +64,7 @@
         "eslint-plugin-prettier": "^4.0.0",
         "eslint-plugin-react": "^7.29.4",
         "husky": "^4.3.0",
+        "is-mobile": "^3.1.1",
```
mxdvl, line 67: I think this becomes an actual dependency, rather than a development one?

Good job on [finding something so tiny](https://bundlephobia.com/package/is-mobile@3.1.1).

mxdvl: We looked into [how `microbundle` decides whether dependencies should be external or inline](https://github.com/developit/microbundle/wiki/How-Microbundle-decides-which-dependencies-to-bundle). For this small a lib (806B minified + gzipped) it makes sense to inline it.

> ## How Microbundle decides inline _vs_ external
> 
> Microbundle decides whether a dependency should be inlined or left as an `import` / `require()` based on how you declare dependencies in your `package.json`.
> 
> When you import a module in your library, Microbundle checks to see how that module was added to your `package.json`.
> 
> - If the module is listed in the `"peerDependencies"` or `"dependencies"` fields, it will be considered _external_ and won't be inlined into your bundled code. External modules remain runtime dependencies of your bundle using `require()` or `import`.
> - Conversely, if the imported module is _only_ referenced in the `"devDependencies"` `package.json` field, that dependency module will be inlined into the bundle.
> 
> This behaviour makes sense when we think about what those fields express: when someone installs your npm package, your listed `"dependencies"` are downloaded too - the expectation is that your package will import or require those modules. The same is true for `"peerDependencies"`, which are automatically downloaded by npm version 7, and produced warnings in earlier npm versions if not already installed.

mxdvl (approved):

package.json
joecowton1: Thanks for adding the explainer Max.

src/YoutubeAtomSticky.tsx
```diff
@@ -104,6 +117,8 @@ type Props = {
     children: JSX.Element;
 };
 
+const isMobile = detectMobile();
```
abeddow91, line 120: Should we include `tablet:true` here so that we detect for mobile *and* tablet given that we want to detect devices with (primarily) touch screens

https://github.com/juliangruber/is-mobile#:~:text=To%20add%20support%20for%20tablets%2C%20set%20tablet%3A%20true.

joecowton1: ah yes - i just merged so will make a new PR to cover that, really well spotted!

src/lib/detectMobile.ts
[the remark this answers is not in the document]
marjisound: Nice one 😄

src/YoutubeAtomSticky.tsx
marjisound: Nice explanation, and thanks for sharing the link 👍

arelra: > Do you think it's possible and beneficial to make isMobile a ref since it should be a static value and doesn't need to change?

Usually the determining factor for using a ref is that it is not related to the render state but you still want to persist or mutate it across renders.

So just to expand on this particular case - iiuc `isMobile` is connected to the render state so shouldn't be a ref.

But refs do have legitimate use cases so it's perfectly fine to use them where they make sense.

> For added context though the React docs do suggest caution when using refs: https://reactjs.org/docs/refs-and-the-dom.html

Just icymi I think the linked docs are outdated as it discusses issues with pre-hooks refs and mainly as handles to dom elements. The React docs are in dire need of an update!

A more up to date reference is here:
https://reactjs.org/docs/hooks-reference.html#useref

Also the (vastly improved) beta of the new React docs are here:
https://beta.reactjs.org/apis/useref

joecowton1: Thanks @arelra -good spot! Also those new beta docs really are infinitely better.

PR #412, opened 2022-05-26
Recipe schema atom

frederickobrien:
This is an attempt to add a barebones recipe atom which simply renders [schema](https://schema.org/Recipe) out of sight for the benefit of web crawlers/search engines. This is part of a proposed experiment to get structured data on a handful of recipe pages and get a before/after view on search engine performance.

We believe all that's needed is a script tag containing a [JSON-LD](https://json-ld.org/) blob. The example used in the atom test has been run through Google's [Rich Results Test](https://search.google.com/test/rich-results) and seems to work fine. 

My thanks to @shtukas for helping me piece this together!

### DIFF
--- a/.changeset/wise-badgers-impress.md
+++ b/.changeset/wise-badgers-impress.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': minor
+---
+
+Adds a barebones recipe schema atom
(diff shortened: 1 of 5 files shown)


shtukas (approved):
+1 from me

mchv:
Just sharing that we already have a defined [model for recipe atoms](https://github.com/guardian/content-atom/blob/master/thrift/src/main/thrift/atoms/recipe.thrift) and that we integrated [them to to web in the past](https://github.com/guardian/frontend/pull/16001).

PR #416, opened 2022-05-31
fix: Add aria labels to like dislike buttons

AshCorr:
## What does this change?

Adds `aria-label` to like and dislike buttons, previously a screen reader would announce these as "button" instead of anything descriptive.

Also makes them easier to click from voice control.

Fixes guardian/dotcom-rendering#5073

## Accessibility

- [x] [Tested with screen reader](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#screen-readers)
- [X] [Navigable with keyboard](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#keyboard-navigation)

### DIFF
--- a/src/expandableAtom/Footer.tsx
+++ b/src/expandableAtom/Footer.tsx
@@ -75,7 +75,7 @@ export const Footer = ({
                 >
                     <div>Was this helpful?</div>
                     <button
-                        aria-label="Like"
+                        aria-label="yes"
                         data-testid="like"
                         css={buttonStyling}
                         onClick={() => {
@@ -86,7 +86,7 @@ export const Footer = ({
                         <ThumbImage />
                     </button>
                     <button
-                        aria-label="Dislike"
+                        aria-label="no"
                         css={[
                             buttonStyling,
                             css`
--- a/src/expandableAtom/Footer.tsx
+++ b/src/expandableAtom/Footer.tsx
@@ -75,6 +75,7 @@ export const Footer = ({
                 >
                     <div>Was this helpful?</div>
                     <button
+                        aria-label="Like"
                         data-testid="like"
                         css={buttonStyling}
                         onClick={() => {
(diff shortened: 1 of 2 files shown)


AshCorr:
I'm not quite sure what feels more natural here, to call these buttons Like/Dislike, Yes/No, or Thumbs up/Thumbs down

My gut reaction is Like/Dislike since thats usually the context of these icons

jlieb10:
Since the question is a "yes" or "no" question, i think the button labels should describe the answer (yes or no) rather than describe the symbols (like or dislike)

jlieb10 (approved):
Approved - but think it should be "yes" or "no" instead of "like" or "dislike" - non-blocking thought

mxdvl (approved):
You’ll want to run `yarn changeset`.

- Yes, this was helpful
- No, this was not helpful

src/expandableAtom/Footer.tsx
```diff
@@ -75,6 +75,7 @@ export const Footer = ({
                 >
                     <div>Was this helpful?</div>
                     <button
+                        aria-label="yes"
```
SiAdcock, line 78:
```suggestion
                        aria-label="yes, this was helpful"
```

src/expandableAtom/Footer.tsx
```diff
@@ -85,6 +86,7 @@ export const Footer = ({
                         <ThumbImage />
                     </button>
                     <button
+                        aria-label="no"
```
SiAdcock, line 89:
```suggestion
                        aria-label="no, this was not helpful"
```

mxdvl (commented):
I like Simon's suggestion and would have proposed the same.

SiAdcock:
Presumably there could be multiple atoms on the page. We should also use `aria-labelled-by` to ensure we link the buttons to this particular atom.

PR #425, opened 2022-06-29
Ash/youtube atom zero

AshCorr:
## What does this change?

Only render the duration overlay on Youtube atoms if duration is not undefined. Sometimes we're unable to get the length of a youtube video and the duration defaults to "0"

### DIFF
--- a/src/YoutubeAtomOverlay.tsx
+++ b/src/YoutubeAtomOverlay.tsx
@@ -135,7 +135,7 @@ export const YoutubeAtomOverlay = ({
                 >
                     <SvgPlay />
                 </div>
-                {duration > 0 && (
+                {duration !== undefined && duration > 0 && (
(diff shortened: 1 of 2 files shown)


src/YoutubeAtomOverlay.tsx
```diff
@@ -135,7 +135,7 @@ export const YoutubeAtomOverlay = ({
                 >
                     <SvgPlay />
                 </div>
-                {duration && (
+                {duration !== undefined && (
```
mxdvl, line 138: I think we would need to make sure that it’s a valid length.

```suggestion
                {duration > 0 && (
```

AshCorr: I'm not sure which scenario it would be less than 0 but I guess this just makes sure

PR #434, opened 2022-07-18
Remove `youtube-player` library

zekehuntergreen:
## What does this change?
- Removes [`youtube-player`](https://www.npmjs.com/package/youtube-player) from the project dependencies and replaces it with 
   - `loadYouTubeIframeApi` function to load the Youtube iframe api script
   -  our own `YouTubePlayer` class which acts as a wrapper over the YouTube Iframe API's player object and "promisifies" the player methods so that they're executed only when the player is ready, much like `youtube-player` does.
- Small change to YoutubeAtom MultipleVideos story so that it includes three different videos 
- Removes 11.2kB / 4.2kB gzip from bundled dependencies

## Why?
- In a subsequent PR we'd like to make a change to `loadYouTubeIframeApi` to enable us to create an experiment to test [interactive media ads](https://developers.google.com/interactive-media-ads) on first-party videos. In order to do this we'll need to modify `loadYouTubeIframeApi` so that it loads an additional script.

## How to test
- [x] Run `yarn storybook` to start storybook and check that the behavior of the `YoutubeAtom` hasn't changed
- [x] link to DCR locally to sense check that youtube players behave as expected
- [ ] release and test DCR in CODE to confirm that players behave as expected

### DIFF
--- a/src/YouTubePlayer.ts
+++ b/src/YouTubePlayer.ts
@@ -10,26 +10,36 @@ type EmbedConfig = {
 type PlayerOptions = YT.PlayerOptions & EmbedConfig;
 
 class YouTubePlayer {
+    id: string;
+    playerOptions: PlayerOptions;
     player: YT.Player | undefined;
+    playerInitialized?: Promise<YT.Player>;
 
     constructor(id: string, playerOptions: PlayerOptions) {
-        loadYouTubeIframeApi().then(() => {
-            // TODO what if this.player used before it's defined?
+        this.id = id;
+        this.playerOptions = playerOptions;
+        this.setup();
+    }
+
+    private async setup() {
+        await loadYouTubeIframeApi();
+        return this.getOrInitializePlayer(this.id, this.playerOptions);
+    }
+
+    private getOrInitializePlayer(id: string, playerOptions: PlayerOptions) {
+        if (this.playerInitialized) {
+            return this.playerInitialized;
+        }
+        this.playerInitialized = new Promise<YT.Player>((resolve) => {
             this.player = new YT.Player(id, playerOptions);
+            resolve(this.player);
         });
+        return this.playerInitialized;
     }
-    on(eventName: string, callBack: unknown) {
-        console.log('on', eventName);
-        return () => {
-            console.log("I'm empty");
-        };
-    }
-    getPlayerState() {
-        console.log(
-            'Eachhashingalgorithmgeneratesthesamenumberofbytesirrespectiveoftheplaintextsize',
-        );
-        return 1;
-        // return this.player.getPlayerState();
+
+    async getPlayerState() {
+        const player = await this.setup();
+        return player.getPlayerState();
     }
 }
 
--- a/src/YouTubePlayer.ts
+++ b/src/YouTubePlayer.ts
@@ -1,4 +1,5 @@
 import type { AdsConfig } from '@guardian/commercial-core';
+import { log } from '@guardian/libs';
 import { loadYouTubeAPI } from './loadYouTubeIframeApi';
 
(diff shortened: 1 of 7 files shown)


arelra (commented):
I paired on this but LGTM 👍 

@joecowton1 @marjisound @oliverlloyd fyi

src/YouTubePlayer.ts
```diff
@@ -0,0 +1,56 @@
...
+            .catch(this.logError);
+    }
+
+    stopVideo(): Promise<void> {
+        return this.playerPromise
+            .then((player) => {
+                player.stopVideo();
+            })
+            .catch(this.logError);
+    }
```
marjisound, line 53: I know you already mentioned you might wanna explore removing the listeners (on component unmount) using `removeEventListener`. Was just wondering if you managed doing that?

marjisound (approved):
It took some time until I could understand the details in this PR. But at the end, I made sense of it all. Thanks @zekehuntergreen and @arelra for walking us through the changes. The PR looks great :) 

Just left a small comment regarding the listener removal just as a question. 

**Below I'm sharing the chat we had earlier for future reference:**

Marjan Kalanaki, 09:19, Edited
Hi all, I understand that we are using the PlayerOptions parameter to add the events like `onReady` and `onStateChange`. Could it be also possible to make use of `addEventListener` on the player to add event listeners instead of using the `PlayerOptions`? And then we could also use `removeEventListener` to remove the listeners? 
My main question is, do you think it makes any difference if we use `PlayerOptions` or  `addEventListener` for adding listeners? or they both are doing the same thing?
If we wanted to use addEventListener, we would need to add a function to our custom YouTubePlayer, right?

Marjan Kalanaki, 09:45
Another question that I'm trying to clear my mind about is, in the old code we were adding `onYouTubeIframeAPIReady` to the Window in the same file where we were creating the player. I'm trying to first understand in the old implementation, when does `onYouTubeIframeAPIReady` get  populated with the player? as soon as the page loads and the player.current is set to YouTubePlayer?

Zeke Hunter-Green, 29 min

Hey Marjan, glad you're enjoying it! Ravi and I were saying the other day we like working on these components as well.
> do you think it makes any difference if we use PlayerOptions or  addEventListener for adding listeners?

I'm pretty sure there isn't a difference between these two ways of registering listeners and the docs seem to suggest that they're equivalent. https://developers.google.com/youtube/iframe_api_reference#Events Since we can define our listeners at the time that we create the player object we thought passing them in as PlayerOptions would be simpler.

Zeke Hunter-Green, 21 min

defining onYouTubeIframeAPIReady on window is part of the functionality that the youtube-player library was handling for us. Their version of loadYouTubeIframeApi.js adds it here: https://github.com/gajus/youtube-player/blob/master/src/loadYouTubeIframeApi.js#L33
Our implementation of loadYouTubeIframeApi is pretty similar in that it loads the youtube api script and then returns a promise that resolves when that script calls window.onYouTubeIframeAPIReady when it's ready.

src/YouTubePlayer.ts
zekehuntergreen: @arelra and I have discovered that `youtube-player` doesn't actually remove event listeners by calling `removeEventListeners` on the underlying player. We think it would be a good idea to add this feature in order to avoid memory leaks but will work on this as a follow up in another PR.

joecowton1 (approved):

src/YouTubePlayer.ts
joecowton1: Sounds good to me.

marjisound: Nice catch, yes sounds good to do it in upcoming PR 👍

PR #443, opened 2022-08-08
YoutubeAtomPlayer removes YouTube event listeners before unmount

zekehuntergreen:
## What does this change?
- Introduces an improvement to the `YoutubeAtomPlayer` component to remove event listeners from the `YT. Player` object before `YoutubeAtomPlayer` unmounts.

## How to test
- [x] run storybook locally. Switching from one YoutubeAtom story to another unmounts the YoutubeAtom component each time.
- [ ] release this change, create a new branch on DCR which upgrades to the new version, deploy this branch to CODE and verify on [a page with multiple Youtube Atoms](https://m.code.dev-theguardian.com/world/live/2022/mar/28/russia-ukraine-war-latest-news-zelenskiy-putin-live-updates) that their listeners are removed when the players unmount.

### DIFF
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -1,5 +1,5 @@
-import React, { useEffect, useRef } from 'react';
-import { YouTubePlayer } from './YouTubePlayer';
+import React, { useEffect, useLayoutEffect, useRef } from 'react';
+import { YouTubeEventListenerName, YouTubePlayer } from './YouTubePlayer';
 
 import type { AdTargeting, VideoEventKey } from './types';
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
@@ -35,6 +35,13 @@ type ProgressEvents = {
     hasSent75Event: boolean;
     hasSentEndEvent: boolean;
 };
+
+type PlayerListenerItem<T extends YouTubeEventListenerName> = {
+    name: T;
+    listener: NonNullable<YT.Events[T]>;
+};
+
+type PlayerListeners = Array<PlayerListenerItem<YouTubeEventListenerName>>;
 
 /**
  * ProgressEvents are a ref, see below
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -36,17 +36,22 @@ type ProgressEvents = {
     hasSentEndEvent: boolean;
 };
 
+/**
+ *  E.g.
+ *  name: onReady, onStateChange, etc...
+ *  listener: YT.PlayerEventHandler<PlayerEvent>, YT.PlayerEventHandler<OnStateChangeEvent>
+ */
 type PlayerListener<T extends PlayerListenerName> = {
-    // onReady, onStateChange, etc.
     name: T;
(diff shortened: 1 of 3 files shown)


src/YoutubeAtomPlayer.tsx
```diff
@@ -367,17 +391,24 @@ export const YoutubeAtomPlayer = ({
...
+    useLayoutEffect(() => {
         /**
-         * Unregister listeners on component unmount
+         * Unregister listeners before component unmount
          *
-         * A useEffect with an empty dependency array will
-         * call its cleanup on unmount and not after every
-         * useEffect update.
+         * A useLayoutEffect with an empty dependency array will
+         * call its cleanup before unmount.
```
zekehuntergreen, line 401: Add comment to explain why useEffect doesn't work

arelra (approved):
LGTM 👍

marjisound:
Thanks @zekehuntergreen for walking me through the PR in a call. We talked about 2 points:
1- From quickly testing it together it seems to be possible to use the existing `playerPromise` for removing the listeners rather than `this._player`. What do you think about this @arelra ?

2- Also for triggering the listener removal at unmount, we did an small experiment where instead of using an array of listeners, we manually removed the listeners one by one. This way, we didn't need to have `ExtractEventType` type and could just use the existing `YT.PlayerEvent` type. Maybe this would reduce the complexity but the downside is we would need to add remove listener for each listener we add.

Interested to hear what your thoughts are :)

marjisound (commented):
Thanks @zekehuntergreen  for walking me through the PR in a call. We talked about 2 points:
1- From quickly testing it together it seems to be possible to use the existing playerPromise for removing the listeners rather than this._player. What do you think about this @arelra ?

2- Also for triggering the listener removal at unmount, we did an small experiment where instead of using an array of listeners, we manually removed the listeners one by one. This way, we didn't need to have ExtractEventType type and could just use the existing YT.PlayerEvent type. Maybe this would reduce the complexity but the downside is we would need to add remove listener for each listener we add.

Interested to hear what your thoughts are :)

marjisound (approved):
Great work @zekehuntergreen 🥳

zekehuntergreen:
Thanks again for the pairing and review @marjisound !
In our test of using `playerPromise` rather than `this._player` to remove event listers from `useLayoutEffect`, I'd missed that this was causing the original warning from the youtube script `The YouTube player is not attached to the DOM`! This must be because the cleanup function returned from `useLayoutEffect` is executed at the right point before the component unmounts but then since `removeEventListener` waits for `playerPromise` to resolve (even though the promise may already be resolved), the callback calling `player.removeEventListener` is put on the call stack and the youtube player is removed from the DOM in the meantime.
We've agreed to leave removing `onReady` and `onStateChange` listeners directly rather than the more generic array of listeners for a possible follow up PR.

marjisound:
Thanks @zekehuntergreen for the update, yes sounds good to me. Lets merge this PR in its current state and later discuss regarding removing the listeners array :)

arelra:
> 2- Also for triggering the listener removal at unmount, we did an small experiment where instead of using an array of listeners, we manually removed the listeners one by one. This way, we didn't need to have ExtractEventType type and could just use the existing YT.PlayerEvent type. Maybe this would reduce the complexity but the downside is we would need to add remove listener for each listener we add.

This is a great simplification thank you @marjisound !

PR #447, opened 2022-08-15
Integrate Interactive Media Ads to YoutubePlayer

zekehuntergreen:
## What does this change?
- introduces an `imaAdTagUrl` param to `YoutubeAtom` and `YoutubeAtomPlayer` as a flag to enable IMA Ads.
- If the flag is set and if the user has consented to targeted ads,
   - Load the IMA SDK script
   - Load the IMA version of the iframe API script
   - include an ad container element adjacent to the player element
   - update Youtube player params
      - disable ads
      - enable IMA
      - disable related videos
   - in Youtube `onReady` callback, initialize an `ImaManager`
- Create refs `imaManager` and `adsManager` which is set in `instantiateImaManager`, so that an IMA ad can be paused when another IMA ad or video on the same page starts playing.
- In `YoutubeAtomPlayer`, move autoplay functionality from the `onReady` callback passed to the `YouTubePlayer` constructor to a `useRef`.

## How to test
- Run storybook locally - the behavior of the YoutubeAtom stories shouldn't have changed.
- link this branch of atoms-rendering to DCR locally, run DCR and frontend, request a page from m.thegulocal.com with video content e.g. https://m.thegulocal.com/news/2022/jul/11/uber-files-whistleblower-lobbyist-mark-macgann

The next step is to create a 0% test on dotcom-rendering in order to test the IMA / PFP integration in production.

## Images
ima demo 480p

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)


src/common/MaintainAspectRatio.tsx
```diff
@@ -24,6 +24,23 @@ export const MaintainAspectRatio = ({
...
+                    height: [pixel value equal to the youtube iframe's height]
+                    position: relative;
+                    display: block;
+                We need to override these styles to make sure that the ad container overlays
+                the youtube player exactly and to avoid a player-sized white space underneath the ad.
+            */
+            [data-atom-type='ima-ad-container'] {
+                width: 100%;
+                height: 100%;
+                position: absolute !important;
```
zekehuntergreen, line 40: Is there a better way than using `!important` here?

src/loadYouTubeIframeApi.ts
```diff
@@ -36,23 +46,28 @@ const youtubeAPIReady = () => {
      * if we're on a subsequent call.
      */
     if (youtubeAPIReadyPromise) {
+        log('dotcom', 'loadYT: returning YTAPI promise');
         return youtubeAPIReadyPromise;
     }
     youtubeAPIReadyPromise = new Promise((resolve) => {
+        log('dotcom', 'loadYT: creating YTAPI promise');
         window.onYouTubeIframeAPIReady = () => {
```
zekehuntergreen, line 54: What if youtube script is ready and calls `window.onYouTubeIframeAPIReady` before this function is defined here?

src/loadYouTubeIframeApi.ts
```diff
@@ -9,18 +9,25 @@ declare global {
...
-    const scripts = [
-        // keep array multi-line
-        loadScript('https://www.youtube.com/iframe_api'),
-    ];
+    let scripts;
+    if (enableIma) {
+        log('dotcom', 'loadYT: loading YT & IMA script');
+        scripts = [
+            loadScript('https://www.youtube.com/iframe_api?ima=1'),
+            loadScript('//imasdk.googleapis.com/js/sdkloader/ima3.js'),
```
zekehuntergreen, line 25: - add comment about allow list
- logging for storybook when run on localhost

src/ImaAdContainer.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+type Props = {
+    adContainerId: string;
+};
+
+const ImaAdContainer = ({ adContainerId }: Props): JSX.Element => {
+    return <div id={adContainerId} data-atom-type="ima-ad-container"></div>;
```
zekehuntergreen, line 8: remove data attribute

src/YoutubeAtomPlayer.tsx
```diff
@@ -221,21 +246,65 @@ const createOnReadyListener =
...
+    (player: YT.Player) => {
+        const makeAdsRequestCallback = (adsRequest: { adTagUrl: string }) => {
+            adsRequest.adTagUrl = imaAdTagUrl;
+        };
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore IMA is an experimental feature and ImaManager is not yet officially part of the YT type
+        if (typeof window.YT.ImaManager !== 'undefined') {
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore see above
+            imaManager.current = new window.YT.ImaManager(
```
zekehuntergreen, line 285: come up with our own types for `ImaManager` to remove any warnings.

zekehuntergreen:
Do we want IMA ads on sticky videos to stick?

src/YoutubeAtomPlayer.tsx
```diff
@@ -9,6 +15,13 @@ import {
     disabledAds,
 } from '@guardian/commercial-core';
 import { log } from '@guardian/libs';
+import type { google } from './ima';
+
+declare global {
+    interface Window {
+        google: typeof google;
```
zekehuntergreen, line 22: why is this done this way??

src/common/MaintainAspectRatio.tsx
```diff
@@ -10,12 +10,11 @@ export const MaintainAspectRatio = ({
     width: number;
     children: React.ReactNode;
 }): JSX.Element => (
-    /* https://css-tricks.com/aspect-ratio-boxes/ */
     <div
         css={css`
             /* position relative to contain the absolutely positioned iframe plus any Overlay image */
             position: relative;
-            padding-bottom: ${(height / width) * 100}%;
+            aspect-ratio: ${width / height};
```
mxdvl, line 17: Have you checked the support for [`aspect-ratio`](https://caniuse.com/mdn-css_properties_aspect-ratio)?

Is this really necessary? If you are not removing the relative positioning, I would keep things as they were.

mxdvl (approved):
Looks good, with two small comments!

src/common/MaintainAspectRatio.tsx
zekehuntergreen: Good call checking the support for aspect-ratio. 
Using `padding-bottom` doesn't work well with the IMA integration since the IMA script changes the [`ima-ad-container`](https://github.com/guardian/atoms-rendering/blob/9266a44016e6fcc74f74d87ec7a908d0c71b2976/src/ImaAdContainer.tsx#L8) styles directly, making it `position: relative` `display: block` and giving it pixel value width and height.
That results in the MaintainAspectRatio component actually appear to have a player-sized padding bottom while the ad is playing.
Screenshot 2022-09-13 at 12 51 11

`aspect-ratio` seemed to be a good way to fix this style issue without having to override the IMA script's styles while also improving readability.

zekehuntergreen: Overriding IMA styles like this seems to work as an alternative:
https://github.com/guardian/atoms-rendering/pull/447/commits/f3fa5634a03ed06e5878f25de9f5e9d0109cbd24

mxdvl: I prefer the override of IMA styles, but you have also made a compelling argument for `aspect-ratio`. I’ll leave it up to you.

zekehuntergreen: I'm happy to stick with padding-bottom in the interest of keeping the changes to users who aren't in the 0% test as light as possible and since as you point out there are quite a few [browsers that don't support aspect ratio](https://caniuse.com/mdn-css_properties_aspect-ratio) in the [list of browsers we officially support](https://www.theguardian.com/help/recommended-browsers).
We can always reconsider when IMA ads are a bit more established.

PR #455, opened 2022-08-29
YoutubeAtom - Pause video rather than stop

arelra:
Co-authored-by: Zeke Hunter-Green <zeke.huntergreen@guardian.co.uk>

## What does this change?

- Update YoutubeAtom behaviour to pause rather than stop when:
  - another video is played on the same page
    (introduced: #401)
  - a sticky video is closed
- Rename `isPlaying` to `isActive` to more closely reflect that the video has started but may be paused. This also helps avoid multiple videos sticking - only an active video can be sticky
- Introduce new state hook `isClosed` to stop a closed video from re-sticking

## Why?
We have done this for a couple of reasons:

- pausing is better aligned to how IMA ads integration (#447) expects the player to behave. Stopping a leads to two state changes, unstarted, and cued which triggers an IMA ad playing.
- better user experience - users won't lose their place in a video when closing it or starting another video.

| | Before      | After      |
|-|------------|------------|
| Close Sticky Video | main close sticky video | pr close sticky video |
| Start second video | main multiple videos |  pr multiple videos |

## How to test

Run storybook with the `Sticky` and `Multiple sticky videos` stories to confirm that 
- when a sticky video is closed, it pauses rather than stops
- a closed video won't stick unless it's played again (becoming active)
- a first video pauses, and if it's sticky, unsticks when a second video is played

### DIFF
--- a/.changeset/wise-lemons-rule.md
+++ b/.changeset/wise-lemons-rule.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': minor
+---
+
+YoutubeAtom - Pause video rather than stop
--- a/src/YouTubePlayer.ts
+++ b/src/YouTubePlayer.ts
@@ -48,6 +48,14 @@ class YouTubePlayer {
             .catch(this.logError);
     }
 
+    pauseVideo(): Promise<void> {
+        return this.playerPromise
+            .then((player) => {
+                player.pauseVideo();
+            })
+            .catch(this.logError);
+    }
+
     stopVideo(): Promise<void> {
         return this.playerPromise
             .then((player) => {
--- a/src/YoutubeAtom.stories.tsx
+++ b/src/YoutubeAtom.stories.tsx
@@ -319,9 +319,9 @@ export const DuplicateVideos = (): JSX.Element => {
     );
 };
 
-export const MultipleVideos = (): JSX.Element => {
+export const MultipleStickyVideos = (): JSX.Element => {
     return (
-        <div style={{ width: '500px' }}>
+        <div style={{ width: '500px', height: '5000px' }}>
             <YoutubeAtom
                 elementId="xyz"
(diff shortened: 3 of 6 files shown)


joecowton1:
@arelra @zekehuntergreen I was trying to remember why we'd reverted to using stop rather than pause when we implemented the sticky videos and found this PR here: https://github.com/guardian/atoms-rendering/pull/401

Have you accounted for that behaviour here?

arelra:
> @arelra @zekehuntergreen I was trying to remember why we'd reverted to using stop rather than pause when we implemented the sticky videos and found this PR here: #401
> 
> Have you accounted for that behaviour here?

Hey @joecowton1 thanks for taking a look.

Yep we did check that the behaviour isn't introduced with this change but we will test further on CODE and have a run through with yourself and @marjisound .

joecowton1:
> > @arelra @zekehuntergreen I was trying to remember why we'd reverted to using stop rather than pause when we implemented the sticky videos and found this PR here: #401
> > Have you accounted for that behaviour here?
> 
> Hey @joecowton1 thanks for taking a look.
> 
> Yep we did check that the behaviour isn't introduced with this change but we will test further on CODE and have a run through with yourself and @marjisound .

Ok perfect, I think that was my only concern. Nice work!

joecowton1 (approved):

marjisound:
The PR is looking great now, while I was testing I just noticed a small thing. When the sticky video is closed manually, it would re-stick again after it goes out of the view. As far as I remember when a user closes the stickiness, it shouldn't stick any more, unless it's in playing state. But @joecowton1 might remember better than me

https://user-images.githubusercontent.com/15894063/187702834-359e8632-19d0-4d9f-92fb-7572cf830461.mov

marjisound (approved):
Amazing work 👏 👏

PR #467, opened 2022-10-13
Add IMA ad label

arelra:
## What does this change?

Enables the IMA YouTube integration UI setting to show an ad label over IMA ads.

## Images

Screenshot 2022-10-13 at 10 59 27

### DIFF
--- a/src/YoutubeAtomPlayer.tsx
+++ b/src/YoutubeAtomPlayer.tsx
@@ -306,6 +306,14 @@ const createInstantiateImaManager =
 
             const onAdsManagerLoaded = () => {
                 adsManager.current = imaManager.current?.getAdsManager();
+                const adsRenderingsettings =
+                    new window.google.ima.AdsRenderingSettings();
+                adsRenderingsettings.uiElements = [
+                    window.google.ima.UiElements.AD_ATTRIBUTION,
+                ];
+                adsManager.current?.updateAdsRenderingSettings(
+                    adsRenderingsettings,
+                );
                 adsManager.current?.addEventListener(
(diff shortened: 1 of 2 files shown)


src/YoutubeAtomPlayer.tsx
```diff
@@ -306,6 +306,14 @@ const createInstantiateImaManager =
 
             const onAdsManagerLoaded = () => {
                 adsManager.current = imaManager.current?.getAdsManager();
+                const adsRenderingsettings =
```
zekehuntergreen, line 309: Do we need to wait to call `updateAdsRenderingSettings` until after the ads manager has loaded? Or could this go directly after instantiating the `imaManager` above?

zekehuntergreen (commented):
Testing locally I'm only seeing the ad label appear some of the time, maybe suggesting a race condition (unless I'm doing something wrong).

arelra:
> I'm only seeing the ad label appear some of the time

Yep good spot. I noticed this too and it seems to be an issue when we have an overlay. I wanted to release this to live so it would be easier for Google to investigate if they wanted to.

src/YoutubeAtomPlayer.tsx
arelra: Once `adsRenderingsettings` is created we set it with `adsManager.current?.updateAdsRenderingSettings(adsRenderingsettings);`
So yes you're right we could create it earlier but I thought it would be a bit tidier to keep it closer to where it was being used.

marjisound (approved):
LGTM 👍

PR #469, opened 2022-10-17
Bump `@guardian/libs@^7.1.0`

arelra:
## What does this change?

Bump `@guardian/libs@^7.1.0` to align types with DCR.

In DCR prop types have diverged since https://github.com/guardian/dotcom-rendering/pull/5750 and tsc fails when bumping atoms-rendering.


marjisound (approved):
Looks good

PR #482, opened 2022-11-01
feat: Use `buildImaAdTagUrl` from commercial-core

arelra:
## What does this change?

We are currently relying on a hard coded ad tag URL passed via props from DCR.

This change now uses `buildImaAdTagUrl` from commercial-core which will build a URL containing all the parameters and targeting required.

Also some minor refactoring:
- `YoutubeAtom` now takes a `imaEnabled` prop. Currently this will only be true if the user if part of the ongoing test.
- move `ImaAdContainer` inside `YoutubeAtomPlayer` (but still rendered as a sibling) as they already share a code path.

## How to test

IMA ads still appear as expected

IMA ads still appear as expected


src/YoutubeAtomPlayer.tsx
```diff
@@ -285,16 +284,25 @@ const createOnReadyListener =
...
         adsManager: React.MutableRefObject<google.ima.AdsManager | undefined>,
     ) =>
     (player: YT.Player) => {
+        const adTargetingEnabled = adTargeting && !adTargeting.disableAds;
+        const adUnit =
+            adTargetingEnabled && adTargeting.adUnit ? adTargeting.adUnit : '';
+        const customParams =
+            adTargetingEnabled && adTargeting.customParams
+                ? adTargeting.customParams
+                : {};
```
arelra, line 300: The defensive logic here is not so nice and a consequence of `adTargeting` being optional.

I think we could tidy this and other sections by making `adTargeting` mandatory - which I think it could be. Will tackle this as a separate PR.

joecowton1 (approved):
Looks good to me @arelra !

PR #491, opened 2022-11-21
Bump source deps and fix missing peer dependencies

joecowton1:
## What does this change?

Updates Source packages to most recent versions and resolves missing peer dependencies.

## Why?

We've notice a number of issues with the quiz atoms - we think this is a consequence of updating Source previously.

However due to an issue with the peer dependencies, these visual bugs have only just been surfaced in Chromatic here but already exist in production. See here for an example: https://www.theguardian.com/football/2022/nov/16/world-cup-quiz-21-tournaments

There is a ticket here to resolve these issues: https://github.com/guardian/source/issues/1620

In the meantime, this PR brings the peer dependancies into line and removes a lot of noise when running `yarn`. We hope it will be easier to catch such errors in future as a result.

co-authored by @sndrs


joecowton1:
@arelra said:

> Thanks for sorting this!
> 
> Looks like there are some valid chromatic diffs.
> 
> I noticed the YoutubeAtom overlay stories are now showing a different font?

@arelra I've moved discussion to here as this was a weird duplicate PR, sorry!

I'm not seeing a different font on the youtube atom though, just text sans. What do you see?

package.json
```diff
@@ -121,31 +122,37 @@
...
         "@emotion/react": "^11.1.5",
-        "@guardian/consent-management-platform": "^10",
-        "@guardian/source-foundations": "^4.0.2",
-        "@guardian/source-react-components": "^4.0.1",
+        "@guardian/commercial-core": "^4.25.0",
+        "@guardian/consent-management-platform": "^10.5.0",
+        "@guardian/libs": "^9.0.1",
+        "@guardian/source-foundations": "^7.0.1",
+        "@guardian/source-react-components": "^9.0.1",
+        "lodash.debounce": "^4.0.6",
```
mxdvl, line 152: Is there a way we could have a script that prevents these ever getting out of sync?

For example, [`discussion-rendering`’s dependencies.ts script does that](https://github.com/guardian/discussion-rendering/blob/18d8f936c450428d0c89f56b6055d5f77cdd7c1f/scripts/dependencies.ts).

mxdvl (approved):
Nice! It would be nice to ensure these things do not get out of sync.

arelra:
> I'm not seeing a different font on the youtube atom though, just text sans. What do you see?

It's very subtle and it may just be a known change in source but in the YoutubeAtom overlay tests the text that displays the time duration of the video is showing a diff. It might be positional rather than the font itself but it is hard to tell.

If this is expected then that's fine just wanted to verify.

Screenshot 2022-11-21 at 17 19 03

package.json
joecowton1: That looks fantastic, are you volunteering @mxdvl ?

PR #498, opened 2022-11-29
Improve accessibility of Quiz atoms

mxdvl:
## What does this change?

Improves the accesibility of our quiz atoms by wrapping the questions in a list using an `ol` and `li` tags.

This helps keyboard and screen-reader users to understand the hierarchy better and move quicker between the questions.

## Why?

Part of https://github.com/guardian/dotcom-rendering/issues/5065

For full fix, this package needs to be updated in `dotcom-rendering`.

### DIFF
--- a/.changeset/twenty-laws-double.md
+++ b/.changeset/twenty-laws-double.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': minor
+---
+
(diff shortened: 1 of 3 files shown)


bryophyta (commented):
I paired on this and it looks good to me 👍

PR #506, opened 2022-12-09
Fix `YoutubeAtom.test` TypeScript errors

arelra:
## What does this change?

Fix `YoutubeAtom.test` TypeScript errors


joecowton1 (approved):
Nice one!

PR #514, opened 2022-12-13
Prefer optional chain fixes

joecowton1:
## What does this change?

Implements eslint rule `prefer-optional-chain`


src/loadYouTubeIframeApi.ts
```diff
@@ -61,9 +61,8 @@ const youtubeAPIReady = () => {
 };
 
 const loadYouTubeAPI = (enableIma = false): Promise<typeof YT> => {
-    // if another part of the code has already loaded youtube api, return early
     /* If another part of the code has already loaded youtube api, return early. */
-    if (window.YT && window.YT.Player && window.YT.Player instanceof Function) {
```
joecowton1, line 66: @zekehuntergreen @arelra was there a reason for this? Typescript / eslint seems fine without the checks.

